PR #17, opened 2015-02-05
Adding synchronous offline logging

philippb:
Hi @curtisliu 
This pull request adds synchronous offline logging.

The main method added is `public static void logEventSynchronous(String eventType, JSONObject eventProperties)` which allows to log an event to amplitude on the same thread as the caller.
This event will not get synced to the server immediately.

This change has been introduced to support logging in a `UncaughtExceptionHandler` so we can see crashes in the users event stream. The reason this needs to be synchronous is that the JVM is already shutting down other threads when handling uncaught exceptions. The event will be pushed to amplitude on next app launch.

We're still in the intense testing phase of this, but I wanted to open this PR to get a first feedback from your side. 

If you change your project to have automated tests, I'm more then happy to add tests to this. Just didn't want to start a refactoring of your SDK project :)

### DIFF
--- a/src/com/amplitude/api/Amplitude.java
+++ b/src/com/amplitude/api/Amplitude.java
@@ -2,6 +2,8 @@ package com.amplitude.api;
 
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
+import java.lang.Exception;
+import java.lang.Thread;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


src/com/amplitude/api/Amplitude.java
```diff
@@ -244,7 +272,8 @@ private static long logEvent(JSONObject event) {
             dbHelper.removeEvents(dbHelper.getNthEventId(Constants.EVENT_REMOVE_BATCH_SIZE));
         }
 
-        if (dbHelper.getEventCount() >= Constants.EVENT_UPLOAD_THRESHOLD) {
+        if (dbHelper.getEventCount() >= Constants.EVENT_UPLOAD_THRESHOLD
+                && !offline) {
```
curtisliu: You'll want to skip the updateServerLater call if offline as well.

curtisliu:
Overall, this looks ok. logEvent(JSONObject event) may need to be synchronized since access to the database is no longer restricted to a single thread.

Also, I'm making some changes to make Amplitude a singleton instead of a bunch of static classes and add a unit test suite, that would help if you wanted to write test cases for this too.

philippb:
@curtisliu Good to see you today.

I made those changes you suggested. Early exit when offline before logging remote and I synchronized the log method. I don't know enough about SQLite to know what it's doing without taking a deep dive.

On the singelton side, have you thought about having a real amplitude class that does all the things, `AmplitudeImpl` with only package visibility, and then your `Amplitude` class that holds on instance and offers all the static methods. Might be nicer to test.

curtisliu:
Cool, if you merge in changes from master you should get exactly what you mentioned -- I've pulled everything into an inner Lib class that Amplitude then delegates to a singleton instance of.

allanca:
It looks like this is no longer relevant. Closing for now.

PR #18, opened 2015-02-19
build.xml: Add target to run the tests

ifrade:
"ant test" compiles and run the unit tests.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -11,6 +11,9 @@
     <property name="build" location="build" />
     <property name="dist" location="dist" />
 
+    <property name="test.src" location="test" />
+    <property name="test.reports" location="${dist}/reports"/>
+
     <property environment="env" />
     <property name="android-sdk-path" value="${env.ANDROID_SDK}" />
     <property name="android-target" value="${env.ANDROID_TARGET}" />
@@ -30,6 +33,7 @@
     <target name="init">
         <mkdir dir="${build}" />
         <mkdir dir="${dist}/lib" />
+        <mkdir dir="${test.reports}" />
     </target>
 
     <target name="clean" >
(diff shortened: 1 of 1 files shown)


build.xml
```diff
@@ -87,6 +91,36 @@
         </artifact:mvn>
     </target>
 
+    <target name="compile-tests" depends="build" description="Compile test files">
```
curtisliu: I think we should name this build-tests

build.xml
```diff
@@ -87,6 +91,36 @@
...
+        <path id="test.classpath">
+            <path refid="project.class.path" />
+            <fileset dir="lib" includes="**/*.jar" />
```
curtisliu: Since the Android JAR just contains stubs and the robolectric classes should take precedence, I think the fileset should come before the path. Same thing below in the test target.

build.xml
```diff
@@ -87,6 +91,36 @@
...
+        <javac srcdir="${test.src}"
+               destdir="${build}"
+               classpathref="test.classpath"
+               includeantruntime="false">
+        </javac>
+    </target>
+
+    <target name="test" depends="compile-tests" description="Runs all Robolectric unit tests.">
+        <path id="junit.classpath">
+            <path refid="project.class.path" />
```
curtisliu: Does this run without the junit4 jar?

curtisliu:
Thanks for this! I think just a few minor changes and I'll merge it in!

build.xml
ifrade: It seems the junit classes are inside the robolectric jar, no need of a external junit jar.

curtisliu:
LGTM!

PR #20, opened 2015-03-04
Let users opt out of event logging

allanca:
An app can now let a user choose whether to be included in tracking. From @curtisliu.


allanca:
LGTM

Issue #16, opened 2015-01-07
getMostRecentLocation causing NullPointerException


berglind:
From our crash logs
```
java.lang.NullPointerException
	at com.amplitude.api.DeviceInfo.getMostRecentLocation(Unknown Source)
	at com.amplitude.api.Amplitude.logEvent(Unknown Source)
                                logEvent
                                logEvent
                                logEvent
	at com.amplitude.api.Amplitude.access$1300(Unknown Source)
	at com.amplitude.api.Amplitude$2.run(Unknown Source)
	at android.os.Handler.handleCallback(Handler.java:605)
	at android.os.Handler.dispatchMessage(Handler.java:92)
	at android.os.Looper.loop(Looper.java:137)
	at android.os.HandlerThread.run(HandlerThread.java:60)
```

allanca:
While I wasn't able to explicitly reproduce the crash, I've added tests for null on the edge cases that will fix this crash. The fix will be included in the next release of the SDK.

PR #21, opened 2015-03-04
Handle null edge cases in location request

allanca:
Location handling has a couple of extreme edge cases that produce null values. Handle those cases to avoid crashing.

Fix for #16.


src/com/amplitude/api/DeviceInfo.java
```diff
@@ -181,7 +181,16 @@ public Location getMostRecentLocation() {
 
         LocationManager locationManager = (LocationManager) context
                 .getSystemService(Context.LOCATION_SERVICE);
+
+        // Don't crash if the device does not have location services.
+        if (locationManager == null) { return null; }
```
curtisliu: style preference, put one liners on a new line anyways

curtisliu:
LGTM

PR #22, opened 2015-03-04
Added instructions for import in gradle files

joninvski:
Included the instruction to import the amplitude library for projects
using gradle, the new build tool recommended by Google to create
android projects.

Signed-off-by: Joao Trindade <trindade.joao@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,37 +15,43 @@
     </dependency>
     ```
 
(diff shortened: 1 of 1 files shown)


allanca:
Thanks, Joao. I'm looking to move the SDK build system to gradle soon as well.

Issue #19, opened 2015-02-27
App Crash on SQLite cursor


philippb:
Amplitude version: `com.amplitude:android-sdk:1.3`

Stack Trace
_________________________________
```
0	android.database.CursorWindowAllocationException: Cursor window allocation of 2048 kb failed. 
1		at android.database.CursorWindow.<init>(CursorWindow.java:108)
2		at android.database.CursorWindow.<init>(CursorWindow.java:100)
3		at android.database.AbstractWindowedCursor.clearOrCreateWindow(AbstractWindowedCursor.java:198)
4		at android.database.sqlite.SQLiteCursor.clearOrCreateWindow(SQLiteCursor.java:301)
5		at android.database.sqlite.SQLiteCursor.fillWindow(SQLiteCursor.java:139)
6		at android.database.sqlite.SQLiteCursor.getCount(SQLiteCursor.java:133)
7		at android.database.AbstractCursor.moveToPosition(AbstractCursor.java:197)
8		at android.database.AbstractCursor.moveToNext(AbstractCursor.java:245)
9		at com.amplitude.api.DatabaseHelper.getEvents(Unknown Source)
10		at com.amplitude.api.Amplitude.updateServer(Unknown Source)
11		at com.amplitude.api.Amplitude.updateServer(Unknown Source)
12		at com.amplitude.api.Amplitude.access$100(Unknown Source)
13		at com.amplitude.api.Amplitude$3.run(Unknown Source)
14		at android.os.Handler.handleCallback(Handler.java:733)
15		at android.os.Handler.dispatchMessage(Handler.java:95)
16		at android.os.Looper.loop(Looper.java:157)
17		at android.os.HandlerThread.run(HandlerThread.java:61)

```

allanca:
Thanks for reporting this.

I've been investigating the crash and though I've been unable to reproduce, I have a few theories about possible causes.

It looks like it occurred on Android 5.0. Do you have any other information about the conditions at the time of the crash? Device type? When in the application lifecycle?

allanca:
Fixed in release 1.4.4 on 3/11.

xiphirx:
Still exists on v1.6.2,

Device types:
![image](https://cloud.githubusercontent.com/assets/993832/7528014/354dbb96-f4d7-11e4-840f-29e61f967872.png)

Android versions:
![image](https://cloud.githubusercontent.com/assets/993832/7528016/3f519450-f4d7-11e4-9375-778c90af93ed.png)

[...]

Issue #9, opened 2014-06-12
ConcurrentModificationException in HashMap


philippb:
```
java.util.ConcurrentModificationException
at java.util.HashMap$HashIterator.nextEntry(HashMap.java:796)
at java.util.HashMap$EntryIterator.next(HashMap.java:833)
at java.util.HashMap$EntryIterator.next(HashMap.java:831)
at org.json.JSONObject.writeTo(JSONObject.java:666)
at org.json.JSONStringer.value(JSONStringer.java:237)
at org.json.JSONObject.writeTo(JSONObject.java:667)
at org.json.JSONObject.toString(JSONObject.java:636)
at com.amplitude.api.Amplitude.logEvent(Unknown Source)
at com.amplitude.api.Amplitude.logEvent(Unknown Source)
at com.amplitude.api.Amplitude.startNewSession(Unknown Source)
at com.amplitude.api.Amplitude.startNewSessionIfNeeded(Unknown Source)
at com.amplitude.api.Amplitude.access$600(Unknown Source)
at com.amplitude.api.Amplitude$4.run(Unknown Source)
at android.os.Handler.handleCallback(Handler.java:587)
at android.os.Handler.dispatchMessage(Handler.java:92)
at android.os.Looper.loop(Looper.java:130)
at android.os.HandlerThread.run(HandlerThread.java:60)
```

philippb:
This is still an issue in version 1.3

![crash distribution](https://photos-2.dropbox.com/t/2/AAB9nQ8cbtz3O1l10p-0wgdXSzGqIVJwDNKPg0EK5lrCLQ/12/620820/png/1024x768/3/1425067200/0/2/Screenshot%202015-02-27%2010.49.51.png/CJTyJSABIAIgAygBKAI/07eubot9Iq4piT6zgHIYgH1IBbXapIxWOUI_AKv4ELQ)

```
0	java.util.ConcurrentModificationException
1		at java.util.HashMap$HashIterator.nextEntry(HashMap.java:792)
2		at java.util.HashMap$EntryIterator.next(HashMap.java:829)
3		at java.util.HashMap$EntryIterator.next(HashMap.java:827)
4		at org.json.JSONObject.writeTo(JSONObject.java:666)
5		at org.json.JSONStringer.value(JSONStringer.java:237)
6		at org.json.JSONObject.writeTo(JSONObject.java:667)
7		at org.json.JSONObject.toString(JSONObject.java:636)
[...]

PR #23, opened 2015-03-04
Fix two crashes

allanca:
Issue #9 (not #16 as reported in the commit) and issue #19 are fixed here.


allanca:
@curtisliu or @paladin8, mind taking a quick look at this one when you get a minute so we can merge and release?

curtisliu:
taking a look!

curtisliu:
LGTM

PR #24, opened 2015-03-05
Automate releases

allanca:
Just change the version number in build.xml and run ```ant release```. It builds, sends all the packages to maven, and commits and tags to git. All that's left is logging into Sonatype and releasing the build.


paladin8:
It looks good, although I'm guessing you haven't been able to actually test the pushing to Sonatype. Let's go over that in person at some point.

allanca:
That's true. I'm around today if you want to try it out together.

PR #27, opened 2015-03-15
Update version number in Constants.java to 1.4.4

daisy1754:
Just noticed version number in Constatns.java is not synced with version number in README


allanca:
Thanks for catching this.

PR #28, opened 2015-03-15
Fix Amplitude.initialize(Context, String, String) to pass around argument properly

daisy1754:


allanca:
Thanks for catching this.

PR #29, opened 2015-03-15
Not mentioning Amplitude.initialize(Context, String, String) in README 

daisy1754:
because it is broken in current release (1.4.4)

see https://github.com/amplitude/Amplitude-Android/pull/28


allanca:
We'll add some tests around your fix and rerelease instead.

PR #30, opened 2015-03-15
Change Amplitude.initialize to throw exception when context or apiKey argument is invalid

daisy1754:
Rather than just logging and swallowing erroneous argument, initialize method should throw exception. There is no use case of passing null context to initialize method.
As a result of this change, contextAndApiKeySet can now simply be isInitialized method.


allanca:
We much prefer not gathering data and logging the error to crashing your app if the sdk is misconfigured for some reason.

I'll see if we can find a way to throw an exception in development mode and swallow the error in release to make it easier catch this case during integration.

PR #31, opened 2015-03-19
SSL pinning

allanca:
Pin the SSL certificate for api.amplitude.com to the Comodo CA certificate. Required changing the http client to use okhttp.

To use SSL pinning, the PinnedAmplitudeClient class must be used.

Also, deprecates the static methods and breaks the client out into a top-level class.


curtisliu:
lgtm! thanks!

PR #33, opened 2015-04-02
syncronous eventLogging

philippb:
Hi @curtisliu 
have a look at this. Took me a while to update that pull request after all your changes.

I had a problem getting all tests to work because of a error
```
[junit] WARNING: multiple versions of ant detected in path for junit
[junit]          jar:file:/usr/local/Cellar/ant/1.9.4/libexec/lib/ant.jar!/org/apache/tools/ant/Project.class
[junit]      and jar:file:/Users/philipp/coding/Amplitude-Android/lib/robolectric-3.0-SNAPSHOT-jar-with-dependencies.jar!/org/apache/tools/ant/Project.class
```

uninstalling ant doesn't help either as I can't run ant :)

Maybe you can have a look at it or let me know whats needed to run.


philippb:
This seems to be the same error then with the test build here:

https://circleci.com/gh/amplitude/Amplitude-Android/35

allanca:
It started happening in the last few days, seemingly unrelated to any code changes. As far as I can tell, a roboelectric library moved off of the maven repository, but I need to dig in a bit deeper to fix it.

The ant warning is a red herring.

-Allan

On April 2, 2015 at 10:19:44 AM, Philipp Berner (notifications@github.com) wrote:

This seems to be the same error then with the test build here:

https://circleci.com/gh/amplitude/Amplitude-Android/35

philippb:
Maybe just a maven `exclude` for the import will solve it.

allanca:
Turns out Robolectric removed some jars for 3.0-SNAPSHOT from the maven repository when they released 3.0-rc2. I just pushed a change that upgrades us to 3.0-rc2 so test will run again.

philippb:
@allanca let me know if you would like some changes or not.

allanca:
I'm not super excited about increasing complexity in the main class with this, though I see the value for logging unhandled exceptions.

We started adding some extra functionality through subclasses recently. This may be another place Would something like this work for you? https://github.com/amplitude/Amplitude-Android/compare/offline-logging

philippb:
I don't think it needs to be in the main app. The important thing is that the log event runs on the same thread as the caller.
For that reason I implemented it offline, so that there is no network traffic on the main thread.

Would we then init another Amplitude offline client?

allanca:
Yes, though maybe the reinitialization would be a problem, since it adds some overhead and device info is populated off the main thread. Could add a copy constructor to OfflineAmplitudeClient... Not sure that's better than finding a way to integrate this simply into the main client, though.

```

philippb:
I'll be at your office on Wed. anyway. How about we talk about this in person and see what we can come up with?

allanca:
Sorry I missed you today.

Thinking more about this, the OfflineAmplitudeClient isn't great either. I think if we take what you have here and break apart the synchronous log method and offline mode, it will be cleaner. Also, turn offline mode into a publicly settable option that works similarly to opt out.

I reworked everything in the offline-logging branch (along with some refactoring I'd been wanting to do) and added some extra tests.

In your exception handler, you would need to both enable offline mode and call the synchronous log method.

Take a look and let me know what you think.

philippb:
Whats your thoughts about this?

allanca:
I think it's in a good place. Take a look at my work at https://github.com/amplitude/Amplitude-Android/compare/offline-logging and let me know what you think.

allanca:
Closing in favor of #40.

PR #37, opened 2015-04-08
Don't use reflection to find api package name.

allanca:
Also, move preference data from the old package name scheme to the new so that no data is lost on upgrades.

### DIFF
--- a/src/com/amplitude/api/AmplitudeClient.java
+++ b/src/com/amplitude/api/AmplitudeClient.java
@@ -94,6 +94,9 @@ public class AmplitudeClient {
             Log.e(TAG, "Argument context cannot be null in initialize()");
             return;
         }
+
+        AmplitudeClient.upgradePrefs(context);
+
         if (TextUtils.isEmpty(apiKey)) {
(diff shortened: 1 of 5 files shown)


src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -779,4 +782,98 @@ private String bytesToHexString(byte[] bytes) {
...
+                    sourcePkgName = Constants.class.getPackage().getName();
+                } catch (Exception e) { }
+            }
+
+            if (targetPkgName == null) {
+                targetPkgName = Constants.PACKAGE_NAME;
+            }
+
+            // No need to copy if the source and target are the same.
+            if (Constants.PACKAGE_NAME.equals(sourcePkgName)) {
```
curtisliu: should this be targetPkgName.equals(sourcePkgName)?

curtisliu:
LGTM, were you able to test the upgrade path manually on an emulator or device?

allanca:
Yes, everything looks correct when moving between versions. I'll merge and ship this as 1.6.0

curtisliu:
Awesome thanks!

PR #39, opened 2015-04-15
Fix var name in revenue logging code.

hfwang:
I had to stop and think because copy/pasting code didn't work :)


allanca:
Great catch. Thanks for the PR.

PR #40, opened 2015-05-04
Offline logging

allanca:
Add support for synchronous logging and offline mode. Useful for logging unhandled exceptions.


curtisliu:
nice, lgtm

PR #41, opened 2015-05-04
Get User Properties and Report Event Logging to Listeners

joecasson:
Hey Curtis + Amplitude dev team, let me know what you think about the following changes!

Adding the functionality to get the user properties stored in Amplitude client would allow for other platforms to be able to use Amplitude's segments. Very helpful for integrations.

Similarly, allowing for other platforms to listen to the events that are being logged by Amplitude. I've included the entire event, but even limiting it to the event type string would be awesome. 

Apologies for the formatting changes in the commit. Changes in AmplitudeClient are around the following lines:
- Line 78
- Line 273
- Line 580 - 585
- Line 900+


haoliu-amp:
Closing this, but we appreciate your contribution in 2015. Thanks @joecasson

PR #43, opened 2015-05-27
Disable Log.DEBUG statements by default.

benjamin-bader:
As ours is a privacy-enhancing app, we'd prefer not to have our event names shown in logcat.

This diff disables debug-level android.Log statements by default, and adds a method to enable them.  Currently there's only one such statement; the rest of the log statements in AmplitudeClient primarily communicate developer errors, and are left on by default here.

### DIFF
--- a/src/com/amplitude/api/AmplitudeClient.java
+++ b/src/com/amplitude/api/AmplitudeClient.java
@@ -51,6 +51,7 @@ public class AmplitudeClient {
     private boolean initialized = false;
     private boolean optOut = false;
     private boolean offline = false;
+    private boolean debugLogEnabled = false;
 
     private DeviceInfo deviceInfo;
     private String advertisingId;
@@ -185,6 +186,10 @@ public class AmplitudeClient {
         }
     }
 
+    public void setDebugLogEnabled(boolean debugLogEnabled) {
+        this.debugLogEnabled = debugLogEnabled;
+    }
+
     public void logEvent(String eventType) {
         logEvent(eventType, null);
     }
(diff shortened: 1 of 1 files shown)


benjamin-bader:
Hm, not sure why the CI build failed ("package org.json does not exist").  Seems unrelated to this diff.

benjamin-bader:
This line from the build log seems suspect, as if some environment variable was missing (or variable substitution is broken):
```
     [echo] Android SDK Path: /home/ubuntu/.m2/repository/com/squareup/okhttp/okhttp/2.2.0/okhttp-2.2.0.jar:/home/ubuntu/.m2/repository/com/squareup/okio/okio/1.2.0/okio-1.2.0.jar:/home/ubuntu/Amplitude-Android/${env.ANDROID_SDK}/platforms/android-${env.ANDROID_TARGET}/android.jar
```
As I'm unfamiliar with both ant and the CircleCI integration, I can't be very helpful here.

benjamin-bader:
Ping

curtisliu:
@benjamin-bader Hi Ben,

I was able to get the CircleCI build to work by changing a project setting. It'd be great if all logging could be enabled/disabled via a helper class (like AmplitudeLog), and even better if you could specify the log level.

benjamin-bader:
Thanks for the good idea.  I'll update soon.

benjamin-bader:
@curtisliu Four months later, this is still a problem for us; any chance that this will land?  If not, I'd prefer that this be closed out.

I think log helpers would be a nice feature, but out-of-scope here.

djih:
We will look into rolling this in for the next release.

djih:
@benjamin-bader : I added a logger helper class: https://github.com/amplitude/Amplitude-Android/commit/d1f94d2ad064122b4a52b2e1508959f9fd2cbca1 the default logging level is Log.INFO, so debug statements won't show up. You can also disable logging and set the logging level via the AmplitudeClient.

benjamin-bader:
Hey, that's awesome - thanks!

PR #46, opened 2015-06-25
Re-use OkHttpClient

niqo01:
* Also enable using a custom OkHttpClient

Jake Wharton mentioned:
```
One OkHttpClient per request is bad. You just lose the shared thread pool and route information.
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,7 @@
 *~
 dist/
 *.asc
+# Intellij
+.idea/
+*.iml
+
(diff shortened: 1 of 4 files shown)


benjamin-bader:
+1, please land this!

djih:
Thanks for bringing this to our attention! We will look into this for our next release.

djih:
We won't allow custom OkHttpClients, but we did fix having more than one instance of the client: https://github.com/amplitude/Amplitude-Android/commit/f04e04d6d44e3d43afa5131ff7b2218e1588bf6d

PR #49, opened 2015-07-17
Daniel/improve android sdk upload logic

djih:
tested the retry logic on Android device as well

### DIFF
--- a/src/com/amplitude/api/AmplitudeClient.java
+++ b/src/com/amplitude/api/AmplitudeClient.java
@@ -72,6 +72,8 @@ public class AmplitudeClient {
     private long eventUploadPeriodMillis = Constants.EVENT_UPLOAD_PERIOD_MILLIS;
     private long minTimeBetweenSessionsMillis = Constants.MIN_TIME_BETWEEN_SESSIONS_MILLIS;
     private long sessionTimeoutMillis = Constants.SESSION_TIMEOUT_MILLIS;
+    private boolean backoffUpload = false;
+    private int backoffUploadBatchSize = eventUploadMaxBatchSize;
 
     private Runnable endSessionRunnable;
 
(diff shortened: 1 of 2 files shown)


src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -703,6 +701,18 @@ public void run() {
...
+            } else if (response.code() == 413) {
+                // If blocked by one massive event, drop it
+                DatabaseHelper dbHelper = DatabaseHelper.getDatabaseHelper(context);
+                if (eventUploadMaxBatchSize == 1) {
+                    dbHelper.removeEvents(maxId);
+                }
+
+                // Server complained about length of request, backoff and try again later
+                int numEvents = Math.min((int)dbHelper.getEventCount(), eventUploadMaxBatchSize);
+                eventUploadMaxBatchSize = (int)Math.ceil(numEvents / 2.0);
```
curtisliu: so eventUploadMaxBatchSize persists until the application falls out of memory, which we don't really have control over. Thinking more I think we want the post request to fully resolve all events up to maxId, and have the request size reduction temporary to a single call of makeEventUploadPostRequest

src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -703,6 +701,18 @@ public void run() {
             } else if (stringResponse.equals("request_db_write_failed")) {
                 Log.w(TAG,
                         "Couldn't write to request database on server, will attempt to reupload later");
+            } else if (response.code() == 413) {
+                // If blocked by one massive event, drop it
+                DatabaseHelper dbHelper = DatabaseHelper.getDatabaseHelper(context);
+                if (eventUploadMaxBatchSize == 1) {
+                    dbHelper.removeEvents(maxId);
```
curtisliu: this will remove all events if the first event is gigantic

djih: It shouldn't since eventUploadMaxBatchSize = 1. It will only fetch 1 event from the dbHelper, which should be the first event. I verified this behavior with my tests and through debugging. I did it this way since I saw the javascript code doing this, although if you want I can explicitly fetch the 0th event and just remove that.

curtisliu: oops you're right!

PR #53, opened 2015-07-29
Daniel/device id sqlite

djih:


src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -948,8 +948,13 @@ static boolean upgradePrefs(Context context, String sourcePkgName, String target
                         source.getLong(sourcePkgName + ".previousSessionId", -1));
             }
             if (source.contains(sourcePkgName + ".deviceId")) {
-                target.putString(Constants.PREFKEY_DEVICE_ID,
-                        source.getString(sourcePkgName + ".deviceId", null));
+                String deviceId = source.getString(sourcePkgName + ".deviceId", null);
```
curtisliu: there needs to be an upgrade path from the current sharedPrefs to the database

src/com/amplitude/api/DatabaseHelper.java
```diff
@@ -22,10 +22,16 @@
     static DatabaseHelper instance;
     private static final String TAG = "com.amplitude.api.DatabaseHelper";
 
+    private static final String AMPLITUDE_STORE_TABLE_NAME = "amplitudestore";
```
curtisliu: how about just "store" or "values"? no need for Amplitude since the database is Amplitude specific

src/com/amplitude/api/DatabaseHelper.java
```diff
@@ -54,10 +61,38 @@ public void onCreate(SQLiteDatabase db) {
 
     @Override
     public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
+        db.execSQL("DROP TABLE IF EXISTS " + AMPLITUDE_STORE_TABLE_NAME);
```
curtisliu: read about SQLiteOpenHelper onUpgrade. You'll need to do version by version upgrading, to preserve old data (we've never really had to use this method before)

src/com/amplitude/api/AmplitudeClient.java
djih: Good catch

src/com/amplitude/api/DatabaseHelper.java
djih: Good point, I like store more

djih:
Added upgrade path from sharedPrefs to db. Fixed db upgrade path. Added some more tests

src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -793,10 +794,19 @@ private String initializeDeviceId() {
...
         invalidIds.add("DEFACE");
 
+        // see if device id already stored in db
+        DatabaseHelper dbHelper = DatabaseHelper.getDatabaseHelper(context);
+        String deviceId = dbHelper.getValue(DEVICE_ID_KEY);
+        if (!TextUtils.isEmpty(deviceId) || invalidIds.contains(deviceId)) {
+            return deviceId;
+        }
+
+        // see if device id stored in old sharedPrefs
```
curtisliu: I think there should be an upgrade path that happens where upgradePrefs happens, and the getDeviceId code should only look in the "current version" store

djih:
Reverted changes to upgradePrefs, added 2nd upgrade path, fixed tests/added new tests for 2nd upgrade path

src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -971,4 +969,35 @@ static boolean upgradePrefs(Context context, String sourcePkgName, String target
...
+
+        String prefsName = sourcePkgName + "." + context.getPackageName();
+        SharedPreferences preferences =
+                context.getSharedPreferences(prefsName, Context.MODE_PRIVATE);
+
+        String deviceId = preferences.getString(Constants.PREFKEY_DEVICE_ID, null);
+        if (!TextUtils.isEmpty(deviceId)) {
+            DatabaseHelper dbHelper = DatabaseHelper.getDatabaseHelper(context);
+            dbHelper.insertOrReplaceKeyValue(DEVICE_ID_KEY, deviceId);
+        }
```
curtisliu: you should remove the source device id afterwards so this transfer only happens once.

djih:
* Removed device id from sharedPrefs after upgrade
* Changed session timeout from 15m to 5m
* Updated changelog and readme to explain new session handling

curtisliu:
LGTM

PR #54, opened 2015-08-12
Concurrent modification exception

djih:
* Following discussion on: https://github.com/candrews/HttpResponseCache/issues/22, best solution is to replace Android implementation of MessageDigest with alternate.
* Using MD5 implementation from: http://org.rodage.com/pub/java/security/MD5.java
* Instrumented following: https://github.com/candrews/HttpResponseCache/commit/eaeaeb2ae4f7905469d4916b8bbc9cf4506b4e30
* Added regression tests that validate new MD5 against Android MessageDigest


test/com/amplitude/security/MD5Test.java
```diff
@@ -0,0 +1,81 @@
...
+        compareMD5(preImage);
+    }
+
+    @Test
+    public void testMD5WithRandomStrings() {
+        for (int i = 0; i < 5; i++) {
+            compareMD5(UUID.randomUUID().toString());
+        }
+    }
+}
```
curtisliu: Add a test for unicode and multi-byte unicode characters (similar to https://github.com/amplitude/Amplitude-Javascript/blob/master/test/md5.js)

djih:
Added unicode tests, ptal

curtisliu:
LGTM

PR #56, opened 2015-08-19
Add target/ directory to .gitignore

friederbluemle:
The Maven build output directory should not show up in `git status`


haoliu-amp:
We're moving to gradle.

PR #57, opened 2015-08-19
Prevent concurrent modification during set user properties

djih:


src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -564,11 +564,20 @@ public void setUserProperties(final JSONObject userProperties, final boolean rep
         runOnLogThread(new Runnable() {
             @Override
             public void run() {
-                Iterator<?> keys = userProperties.keys();
+                // Create deep copy of input userProperties to prevent ConcurrentModification
+                final JSONObject copy;
+                try {
+                    copy = new JSONObject(userProperties.toString());
```
curtisliu: Isn't there still a risk of ConcurrentModification cause you're calling toString here?

djih:
Updated to make deep copy of input JSONObject, PTAL. Let me know if you want to catch or raise the ConcurrentModificationException

src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -548,12 +548,24 @@ public void setUserProperties(JSONObject userProperties) {
...
+        if (userProperties == null) {
+            if (replace) {
+                this.userProperties = null;
+            }
             return;
         }
 
-        if (userProperties == null) {
+        if (replace || this.userProperties == null) {
+            // Only create copy on main thread if necessary
```
curtisliu: Thinking about it the entire replacement should happen on the LogThread

djih: I was originally thinking about that too, but I was worried about the capture in the run. getInstance() grabs the wrong instance, but looks like I can capture "this" for the run.

djih:
Moved all copying onto background thread, PTAL

djih:
Moved all logic onto background thread, PTAL

curtisliu:
LGTM

PR #58, opened 2015-08-21
Catch NullPointerExceptions and ResourceNotFoundExceptions during getCountry

djih:


src/com/amplitude/api/DeviceInfo.java
```diff
@@ -145,11 +148,17 @@ private String getCountryFromLocation() {
...
-                if (country != null) {
-                    return country.toUpperCase(Locale.US);
+            try {
+                if (manager.getPhoneType() != TelephonyManager.PHONE_TYPE_CDMA) {
+                    String country = manager.getNetworkCountryIso();
+                    if (country != null) {
+                        return country.toUpperCase(Locale.US);
+                    }
                 }
+            } catch (NullPointerException e) {
```
curtisliu: I hesitate to say this, but I think catching all Exceptions would make sense here, as it's a system call and not much we can do about it. It should probably also include the getSystemService call on line 149.

djih:
Catch all exceptions from telephonymanager, PTAL

curtisliu:
LGTM

PR #60, opened 2015-09-17
User property operations

djih:
1. Added identify class + tests
2. Added identifys table to DatabaseHelper + tests
3. Updated AmplitudeClient to use identifys table + truncate string values
4. Fixed thread post enqueue in reverse order issue


djih:
Added a key, long value table to store a shared sequence number for merging events and identifys. Tested on Android demo, PTAL @curtisliu

src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -285,13 +290,13 @@ protected void logEventAsync(final String eventType, JSONObject eventProperties,
...
-                logEvent(eventType, copyEventProperties, apiProperties, timestamp, outOfSession);
+                logEvent(eventType, copyEventProperties, apiProperties, userProperties, timestamp, outOfSession);
             }
         });
     }
 
-    protected long logEvent(String eventType, JSONObject eventProperties,
-            JSONObject apiProperties, long timestamp, boolean outOfSession) {
+    protected long logEvent(String eventType, JSONObject eventProperties, JSONObject apiProperties,
+                            JSONObject userProperties, long timestamp, boolean outOfSession) {
```
curtisliu: [style] 8 space continuation indents

src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -550,53 +593,91 @@ public void logRevenue(String productId, int quantity, double price, String rece
...
-                    Log.e(TAG, e.toString());
-                    return; // could not create copy, cannot merge
-                } // catch (ConcurrentModificationException e) {}
-
-                JSONObject currentUserProperties = instance.userProperties;
-                if (replace || currentUserProperties == null) {
-                    instance.userProperties = copy;
-                    return;
+                // create shallow copy to try and prevent ConcurrentModificationException
+                JSONObject copy = cloneJSONObject(userProperties);
```
curtisliu: what's the motivation behind switching deep copy to shallow copy?

src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -550,53 +593,91 @@ public void logRevenue(String productId, int quantity, double price, String rece
...
+            }
+        }
+
+        return object;
+    }
+
+    public JSONArray truncate(JSONArray array) throws JSONException {
+        for (int i = 0; i < array.length(); i++) {
+            Object value = array.get(i);
+            if (value.getClass().equals(String.class)) {
```
curtisliu: is it possible for value to be null here?

src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -676,9 +764,48 @@ public void run() {
...
     }
 
-    protected void makeEventUploadPostRequest(OkHttpClient client, String events, final long maxId) {
+    protected Pair<Pair<Long,Long>, JSONArray> mergeEventsAndIdentifys(List<JSONObject> events,
+                            List<JSONObject> identifys, int numEvents) throws JSONException {
+        JSONArray merged = new JSONArray();
+        long maxEventId = -1;
+        long maxIdentifyId = -1;
+
+        while (merged.length() < numEvents) {
```
curtisliu: again, this sorting I think deserves it's own method

src/com/amplitude/api/DatabaseHelper.java
```diff
@@ -53,43 +62,74 @@ private DatabaseHelper(Context context) {
...
+        switch (oldVersion) {
             case 1:
                 db.execSQL(CREATE_STORE_TABLE);
+                if (newVersion <= 2) break;
 
             case 2:
-                break;
+                db.execSQL(CREATE_IDENTIFYS_TABLE);
+                db.execSQL(CREATE_LONG_STORE_TABLE);
+        if (newVersion <= 3) break;
```
curtisliu: indentation seems a little weird here

src/com/amplitude/api/AmplitudeClient.java
djih: Fixed

src/com/amplitude/api/AmplitudeClient.java
djih: I noticed logEventAsync was shallow copying the event properties JSON, so I thought it might make sense to do the same for user properties. But now that I think about it, clients are probably more likely to maintain a persistent user property JSON object. Switching back to deep copy.

src/com/amplitude/api/AmplitudeClient.java
djih: Good call, will add null checks and tests

src/com/amplitude/api/AmplitudeClient.java
djih: It is in a new method: mergeEventsAndIdentifys, which is called by updateServer

src/com/amplitude/api/DatabaseHelper.java
djih: Fixed

djih:
Added null check to truncate, fixed some indentation issues, PTAL @curtisliu

src/com/amplitude/api/Identify.java
```diff
@@ -0,0 +1,73 @@
...
+
+    public Identify unset(String property) {
+        addToUserProperties(Constants.AMP_OP_UNSET, property, "-");
+        return this;
+    }
+
+    private void addToUserProperties(String operation, String property, Object value) {
+        // check if property already used in previous operation
+        if (userProperties.contains(property)) {
+            Log.e(TAG, String.format("Already used property %s in previous operation, ignoring operation %s", property, operation));
```
curtisliu: Make this a warning

curtisliu:
LGTM

PR #61, opened 2015-09-17
Don't fetch device ID on main thread

djih:
This is caused by onEnterForeground which can send start/end session events on main thread. The fix is to start new session if necessary on the background thread. Start new session should still log events synchronously (in the case where logging an event starts a new session, we want the start/end session events to be sent before the logged event).

It should be okay to toggle the inForeground boolean on the background thread, since log event is the only function that uses it, and that should be called on the background thread.

### DIFF
--- a/src/com/amplitude/api/AmplitudeClient.java
+++ b/src/com/amplitude/api/AmplitudeClient.java
@@ -502,9 +502,14 @@ public class AmplitudeClient {
         logEvent(sessionEvent, null, apiProperties, timestamp, false);
     }
 
-    void onExitForeground(long timestamp) {
-        refreshSessionTime(timestamp);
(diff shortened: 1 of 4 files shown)


src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -507,9 +507,14 @@ void onExitForeground(long timestamp) {
         inForeground = false;
```
curtisliu: probably run this block onLogThread too

curtisliu:
LGTM

src/com/amplitude/api/AmplitudeClient.java
djih: Done

PR #68, opened 2015-10-16
add logger helper class

djih:
tested on Android demo, disabling works to silence log statements


src/com/amplitude/api/AmplitudeLog.java
```diff
@@ -0,0 +1,106 @@
...
+        if (enableLogging) return Log.w(tag, tr);
+        return 0;
+    }
+
+    int w(String tag, String msg, Throwable tr) {
+        if (enableLogging) return Log.w(tag, msg, tr);
+        return 0;
+    }
+
+    int wtf(String tag, String msg) {
```
curtisliu: wtf

src/com/amplitude/api/AmplitudeLog.java
```diff
@@ -0,0 +1,106 @@
...
+    protected static AmplitudeLog instance = new AmplitudeLog();
+
+    public static AmplitudeLog getLogger() {
+        return instance;
+    }
+
+    private AmplitudeLog() {
+    } // prevent instantiation
+
+    void setEnableLogging(boolean enableLogging) {
```
curtisliu: would prefer this to be a setLogLevel

curtisliu: actually leave this and add a setLogLevel

src/com/amplitude/api/AmplitudeLog.java
djih: lol are you saying wtf to the wtf method. That's a legit method: http://developer.android.com/reference/android/util/Log.html#wtf(java.lang.String, java.lang.Throwable)   it stands for What a Terrible Failure

curtisliu: hah

src/com/amplitude/api/AmplitudeLog.java
djih: what is setLogLevel? That's not supported by the Android logger

djih:
Added logLevel PTAL @curtisliu

curtisliu:
LGTM

PR #70, opened 2015-11-24
Log if Google Play Services enabled in api properties['gps_enabled']

djih:
The standard way to check if GPS enabled is this:
```
int statusCode = GooglePlayServicesUtil.isGooglePlayServicesAvailable(context);
if (statusCode == ConnectionResult.SUCCESS) {
    ...
}
```

Note: You will notice that on Line 212 I do statusCode == 0, instead of ConnectionResult.SUCCESS, because ConnectionResult is another gms library and I don't want to load that through reflection just to get the value of [success, which is just 0 anyways](https://developers.google.com/android/reference/com/google/android/gms/common/ConnectionResult.html#SUCCESS).

I do have a test (testGPSEnabled), which does use ConnectionResult.SUCCESS, which will break and let us know if Google ever decides to change the mapped value. I also tested it on the Android demo (with GPS enabled and disabled).

### DIFF
--- a/src/com/amplitude/api/DeviceInfo.java
+++ b/src/com/amplitude/api/DeviceInfo.java
@@ -291,7 +291,7 @@ public class DeviceInfo {
         return getCachedInfo().limitAdTrackingEnabled;
     }
 
-    public boolean isGPSEnabled() { return getCachedInfo().gpsEnabled; }
+    public boolean isGooglePlayServicesEnabled() { return getCachedInfo().gpsEnabled; }
 
     public Location getMostRecentLocation() {
         if (!isLocationListening()) {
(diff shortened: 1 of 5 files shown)


src/com/amplitude/api/DeviceInfo.java
```diff
@@ -262,6 +291,8 @@ public boolean isLimitAdTrackingEnabled() {
         return getCachedInfo().limitAdTrackingEnabled;
     }
 
+    public boolean isGPSEnabled() { return getCachedInfo().gpsEnabled; }
```
curtisliu: `isGooglePlayServicesEnabled()`

(GPS could be confused with Global Positioning System)

curtisliu:
LGTM

src/com/amplitude/api/DeviceInfo.java
djih: Changed

PR #75, opened 2016-02-03
Allow multiple apps

djih:
Allowing multiple apps on Android SDK is way more straightforward than on JS SDK. 

Android SDK is already setup with Amplitude / AmplitudeClient classes, with Amplitude.getInstance() a standard part of our API. Just need to update Amplitude to manage a map of different instance names to different AmplitudeClient instances. Like with JS SDK, if the instance name is blank or null, then we default to a "$defaultInstance" name.

All events and a few meta data + settings are stored in a Sqlite db file. To separate these out for each instance, we give each instance its own Sqlite db file, and the name of the db file is appended with that instance's apiKey for uniqueness. This does mean that the AmplitudeClient instance needs to be initialized with the apiKey before being able to interact with the database. We already do an api key check in all public methods. 

DatabaseHelper has been updated to also manage separate instances for each Sqlite db file. Here we simply use the api key as the map key.

Apps upgrading from single API key to multiple API keys will probably want to keep continuity with their existing apps (keep the same deviceId/userIds for returning users). Added a migration path where we make a copy of the old db file, and copy it to a new db file with the new naming scheme. For people who want a completely new blank app (not inheriting from existing app), I added a flag during initialization to skip the migration process.

It looks like there are a lot of deletions and additions in this PR, but most of it is just migrating tests from AmplitudeTests to AmplitudeClientTests.

### DIFF
--- a/src/com/amplitude/api/AmplitudeClient.java
+++ b/src/com/amplitude/api/AmplitudeClient.java
@@ -19,6 +19,7 @@ import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
 import java.security.MessageDigest;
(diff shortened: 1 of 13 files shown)


src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -906,7 +915,10 @@ protected void makeEventUploadPostRequest(OkHttpClient client, String events, fi
                 logThread.post(new Runnable() {
                     @Override
                     public void run() {
-                        DatabaseHelper dbHelper = DatabaseHelper.getDatabaseHelper(context);
+                        DatabaseHelper dbHelper = DatabaseHelper.getDatabaseHelper(
+                            context,
+                            apiKeySuffix
```
djih: I could probably combine this into one line

PR #76, opened 2016-02-03
Allow multiple apps v2

djih:
Android SDK is already setup with Amplitude / AmplitudeClient classes, with Amplitude.getInstance() a standard part of our API. Just need to update Amplitude to manage a map of different instance names to different AmplitudeClient instances. Like with JS SDK, if the instance name is blank or null, then we default to a "$defaultInstance" name.

For backwards compatibility, all of the existing data, settings, deviceId, any unsent events will remain with the default instance. This provides the smoothest transition as most people won't event have to change their code. Nothing will change at all for people who don't need multiple apps.

All events and a few meta data + settings are stored in a Sqlite db file. To separate these out for each instance, we give each instance its own Sqlite db file, and the name of the db file is appended with that instance's name. This does mean that if they change the instance name, they will lose all their data. I made this very clear in the Readme.

DatabaseHelper has been updated to also manage separate instances for each Sqlite db file. Here we simply use the instance name as the map key.

If people want to synchronize the deviceId between different apps, they can do so via `getDeviceId` and `setUserId`. I explained this in the Readme as well. We explicitly leave this to the user to do, and don't assume anything during the initialization process. This removes the need for any data migrations.

It looks like there are a lot of deletions and additions in this PR, but most of it is just migrating tests from AmplitudeTests to AmplitudeClientTests.


src/com/amplitude/api/DatabaseHelper.java
```diff
@@ -46,20 +51,43 @@
...
+                dbHelper = new DatabaseHelper(context.getApplicationContext());
+            } else {
+                dbHelper = new DatabaseHelper(context.getApplicationContext(), instance);
+            }
+            instances.put(instance, dbHelper);
+        }
+        return dbHelper;
     }
 
+    // Maintain backwards compatability, allow access to old original database file
```
paladin8: Since the method is private, this is not actually for backwards compatibility right? It's for testing? (And the comment probably makes more sense on the static method above.)

test/com/amplitude/api/AmplitudeTest.java
```diff
@@ -53,1073 +31,116 @@ public void tearDown() throws Exception {
...
+        Amplitude.instances.clear();
+        DatabaseHelper.instances.clear();
+
+        String newInstance1 = "newApp1";
+        String newApiKey1 = "1234567890";
+        String newInstance2 = "newApp2";
+        String newApiKey2 = "0987654321";
+        File oldDbFile = context.getDatabasePath(Constants.DATABASE_NAME);
+        File newDbFile1 = context.getDatabasePath(Constants.DATABASE_NAME + "_" + newInstance1);
+        File newDbFile2 = context.getDatabasePath(Constants.DATABASE_NAME + "_" + newInstance2);
```
paladin8: This can probably go either way, but it might be cleaner to have a ```DatabaseHelper``` reference expose an existence check instead of having manually checking the file?

paladin8:
lgtm

src/com/amplitude/api/DatabaseHelper.java
djih: It is for backwards compatibility. The default instance needs access to the original db file. Maybe I shouldn't mark it as deprecated. And yea it is private so maybe not necessary

test/com/amplitude/api/AmplitudeTest.java
djih: Good idea, much cleaner

PR #77, opened 2016-02-08
Update gitignore and SDK path env var

friederbluemle:
Adds four entries to .gitignore that should not be committed to source control, but will show up in `git status` after building the project or opening it in Eclipse.
Also makes a small change to the SDK path environment variable in build.xml. The standard name is `ANDROID_HOME` - no need for a duplicate "`ANDROID_SDK`".

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,7 @@ dist/
 *.asc
 .idea/
 *.iml
+.classpath
+.project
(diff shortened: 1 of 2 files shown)


haoliu-amp:
Thanks for your change. We'll be changing to gradle. I am closing this.

PR #81, opened 2016-02-29
Add Wrapper to Un-Set User Properties

wtowns:
Adds ``unsetUserProperties`` to AmplitudeClient and AmplitudePlugin.  This is meant to make it much easier for Unity to access this functionality.


wtowns:
Any idea what the CircleCI failure is about?  Doesn't necessarily seem related to this change.  Tests are working on my machine at least.

PR #83, opened 2016-03-15
V2.5.1 patch

djih:
Issue: infinite threading loop in the case where a successful event upload triggers another event upload of the remaining events. The second call to updateServer will have limit set to false, which sets the batchLimit to -1. This is so that the dbHelper knows to fetch all remaining events and identifies; however, this breaks the mergeEvent logic, resulting in 0 events passed to makeEventUploadRequest. This doesn't cause the SDK/app to run out of memory since we are always posting to the log and http threads. However, when the app is idling, these 2 threads will continue to call the functions without stop.

Changes: always have batchLimit be the default (30). The limit argument in updateServer just switches between using the default limit or using the backoffUploadLimit (for the backoff). Return out of updateServer if there are no events, instead of just always posting to the http thread. 

Added a test case for this.

Pushing this as a v2.5.1 patch for immediate release, will then merge into master

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Unreleased
 
+* Fix bug where updateServer sets the wrong batchLimit when limit is false.
+
 ## 2.5.0 (January 15, 2016)
 
 * Add ability to clear all user properties.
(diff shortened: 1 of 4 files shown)


paladin8:
lgtm

djih:
Released in v2.5.1: https://github.com/amplitude/Amplitude-Android/releases/tag/v2.5.1

Issue #88, opened 2016-03-23
App crashes with IndexOutOfBoundsException


ipereziriarte:
Hi,
I've updated amplitude to version 2.5.1 and I'm getting a lot of crashes. Here's the stacktrace:

> Fatal Exception: java.lang.IndexOutOfBoundsException
       at java.util.LinkedList.remove(LinkedList.java:660)
       at com.amplitude.api.AmplitudeClient.mergeEventsAndIdentifys(AmplitudeClient.java:844)
       at com.amplitude.api.AmplitudeClient.updateServer(AmplitudeClient.java:816)
       at com.amplitude.api.AmplitudeClient.updateServer(AmplitudeClient.java:789)
       at com.amplitude.api.AmplitudeClient$7.run(AmplitudeClient.java:783)
       at android.os.Handler.handleCallback(Handler.java:739)
       at android.os.Handler.dispatchMessage(Handler.java:95)
       at android.os.Looper.loop(Looper.java:145)
       at android.os.HandlerThread.run(HandlerThread.java:61)

Can you help me please?

djih:
Hi @ipereziriarte sorry for the crashes. Are you able to reproduce the crash yourself? Any other details you can share? We are working on a patch now

djih:
@ipereziriarte does your app run on multiple processes?

ipereziriarte:
Yes. It the app runs in 2 processes

djih:
Ok I think I know what the problem is. We use Android Sharedpreferences to store some meta data (like event id, identify id), and it's known to not be process-safe (http://stackoverflow.com/questions/4693387/sharedpreferences-and-thread-safety).

During updateServer, we fetch the last eventId and last identifyId, and use them to fetch events to upload. I can imagine a scenario where the last eventId or last identifyId value fetched is out of date, but the total event count is correct. The number of events the SDK can merge is less than the number it's expecting to merge, and thus triggers the out of bounds. The solution is to migrate the meta data out of SharedPrefs into the db.

djih:
@ipereziriarte the fix is merged, we will release a new version either tomorrow or early next week

aminelaadhari:
We had the issue but with a single process

djih:
@aminelaadhari  good to know! The patch should fix it for single processes as well.

djih:
[...]

PR #89, opened 2016-03-24
Merging events bug

djih:
We use Android Sharedpreferences to store some meta data (like event id, identify id), and it's known to not be process-safe (http://stackoverflow.com/questions/4693387/sharedpreferences-and-thread-safety).

During updateServer, we fetch the last eventId and last identifyId, and use them to fetch events to upload. I can imagine a scenario where the last eventId or last identifyId value fetched is out of date, but the total event count is correct. The number of events the SDK can merge is less than the number it's expecting to merge, and thus triggers the out of bounds. The solution is to migrate the meta data out of SharedPrefs into the db.

Addresses: https://github.com/amplitude/Amplitude-Android/issues/88


src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -111,15 +120,15 @@ public synchronized AmplitudeClient initialize(Context context, String apiKey, S
...
             this.sharedPreferencesName = getSharedPreferencesName();
             initializeDeviceInfo();
-            SharedPreferences preferences = context.getSharedPreferences(
-                    this.sharedPreferencesName, Context.MODE_PRIVATE);
+
+            DatabaseHelper dbHelper = DatabaseHelper.getDatabaseHelper(this.context);
             if (userId != null) {
-                this.userId = userId;
-                preferences.edit().putString(Constants.PREFKEY_USER_ID, userId).commit();
+                dbHelper.insertOrReplaceKeyValue(USER_ID_KEY, userId);
```
djih: Need to set this.userId = userId as well

src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -111,15 +120,15 @@ public synchronized AmplitudeClient initialize(Context context, String apiKey, S
...
             } else {
-                this.userId = preferences.getString(Constants.PREFKEY_USER_ID, null);
+                this.userId = dbHelper.getValue(USER_ID_KEY);
             }
-            this.optOut = preferences.getBoolean(Constants.PREFKEY_OPT_OUT, false);
+            Long optOut = dbHelper.getLongValue(OPT_OUT_KEY);
+            this.optOut = optOut != null && optOut == 1;
```
djih: Add tests to verify loading from db long -> bool conversion works correctly

src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -874,14 +862,19 @@ protected void updateServer(boolean limit) {
 
                 final Pair<Pair<Long, Long>, JSONArray> merged = mergeEventsAndIdentifys(
                         events, identifys, batchSize);
+                final JSONArray mergedEvents = merged.second;
+                if (mergedEvents.length() == 0) {
+                    uploadingCurrently.set(false);
```
djih: Shouldn't need this condition

src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -898,14 +891,21 @@ public void run() {
         long maxIdentifyId = -1;
 
         while (merged.length() < numEvents) {
+            boolean noEvents = events.isEmpty();
+            boolean noIdentifys = identifys.isEmpty();
+
+            // case 0: no events or identifys, nothing to grab
+            if (noEvents && noIdentifys) {
+                break;
```
djih: note that this case should never occur. if it does then it is an error, and should log a warning

PR #92, opened 2016-03-29
Set Groups

djih:
Reverts amplitude/Amplitude-Android#91


paladin8:
Same comment about multiple groups.

PR #98, opened 2016-04-19
Revenue v2

djih:
Add a new `Revenue` class that wraps revenue properties. The new class adds support for new fields like `revenueType`, logging event properties with revenue events, etc. Contains a little bit of logic to do some verification (like to make sure all the required fields are present). Converts into an event_properties dictionary which can be passed into logEvent


djih:
Finished testing in Android demo app. Revenue events are showing up in the beta revenue tab. Revenue props in event props do not get processed into revenue tab for some reason (filed a ticket for it). But we can still push logRevenueV2 in the meantime. No sdk code changes are needed to fix revenue props in event props.

PTAL @paladin8

README.md
```diff
@@ -239,25 +239,48 @@ Amplitude.getInstance().clearUserProperties();
 
 # Tracking Revenue #
 
-To track revenue from a user, call `logRevenue()` each time a user generates revenue. For example:
+The preferred method of tracking revenue for a user now is to use `logRevenueV2()` in conjunction with the provided `Revenue` interface. `Revenue` instances will store each revenue transaction and allow you to define several special revenue properties (such as revenueType, productId, etc) that are used in Amplitude dashboard's Revenue tab. You can now also add event properties to the revenue event, via the revenueProperties field. These `Revenue` instance objects are then passed into `logRevenueV2` to send as revenue events to Amplitude servers. This allows us to automatically display data relevant to revenue on the Amplitude website, including average revenue per daily active user (ARPDAU), 1, 7, 14, 30, 60, and 90 day revenue, lifetime value (LTV) estimates, and revenue by advertising campaign cohort and daily/weekly/monthly cohorts.
```
paladin8: I don't know what our general philosophy is on this, but it feels like this is more detail than necessary for SDK documentation. Specifically, when we change dashboard functionality I don't want to have to remember to update SDK docs as well. Can we just link to the actual revenue docs?

README.md
```diff
@@ -239,25 +239,48 @@ Amplitude.getInstance().clearUserProperties();
...
+Each time a user generates revenue, you create a `Revenue` object and fill out the revenue properties:
+```java
+Revenue revenue = new Revenue().setProductId("com.company.productId").setPrice(3.99).setQuantity(3);
+Amplitude.getInstance().logRevenueV2(revenue);
+```
 
-**To enable revenue verification, copy your Google Play License Public Key into the manage section of your app on Amplitude. You must put a key for every single app in Amplitude where you want revenue verification.**
+`productId`, `price`, and `quantity` are required fields (`quantity` defaults to 1 if unspecified). `receipt` and `receiptSignature` are required if you want to verify the revenue event. Each field has a corresponding `set` method (for example `setProductId`, `setQuantity`, etc). This table describes the different fields available:
```
paladin8: Quantity is not required if it defaults to 1.

README.md
```diff
@@ -239,25 +239,48 @@ Amplitude.getInstance().clearUserProperties();
...
+| receipt            | String     | Optional: required if you want to verify the revenue event                                               | null    |
+| receiptSignature   | String     | Optional: required if you want to verify the revenue event                                               | null    |
+| revenueProperties  | JSONObject | Optional: a JSONObject of event properties to include in the revenue event                               | null    |
 
-```java
+Note: the price can be negative, which might be useful for tracking revenue lost, for example refunds or costs.
+
+### Revenue Verification ###
 
+By default Revenue events recorded on the Android SDK appear in Amplitude dashboards as unverified revenue events. **To enable revenue verification, copy your Google Play License Public Key into the manage section of your app on Amplitude. You must put a key for every single app in Amplitude where you want revenue verification.**
```
paladin8: Again, linking to the docs seems better than repeating information.

src/com/amplitude/api/Utils.java
```diff
@@ -0,0 +1,47 @@
...
+    static JSONObject cloneJSONObject(final JSONObject obj) {
+        if (obj == null) {
+            return null;
+        }
+
+        // obj.names returns null if the json obj is empty.
+        JSONArray nameArray = null;
+        try {
+            nameArray = obj.names();
+        } catch (ArrayIndexOutOfBoundsException e) {
```
paladin8: Huh, is this possible?

paladin8:
lgtm

README.md
djih: I kind of just assume that some people only look at SDK readmes instead of the docs

README.md
djih: Good catch

src/com/amplitude/api/Utils.java
djih: No idea, but we've been using it for a while now. I think the other way to clone JSONObjects is to do like a new JSONObject(old.toString()) or something, which is why this method was added

Issue #97, opened 2016-04-19
NullPointerException when using PinnedAmplitudeClient with enableForegroundTracking()


dmnugent80:
We're getting a NullPointerException when trying to use PinnedAmplitudeClient with a call to enable foreground tracking:

```
PinnedAmplitudeClient.getInstance().enableForegroundTracking(this);
```

I verified that the same crash occurs if you take the tilt maze demo, and switch it to PinnedAmplitudeClient.

stack trace:

```
04-18 17:51:39.859 E/AndroidRuntime( 4012): FATAL EXCEPTION: logThread
04-18 17:51:39.859 E/AndroidRuntime( 4012): Process: com.example.www, PID: 4012
04-18 17:51:39.859 E/AndroidRuntime( 4012): java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.Long com.amplitude.api.DatabaseHelper.getLongValue(java.lang.String)' on a null object reference
04-18 17:51:39.859 E/AndroidRuntime( 4012): 	at com.amplitude.api.AmplitudeClient.getLongvalue(AmplitudeClient.java:453)
04-18 17:51:39.859 E/AndroidRuntime( 4012): 	at com.amplitude.api.AmplitudeClient.getLastEventTime(AmplitudeClient.java:466)
04-18 17:51:39.859 E/AndroidRuntime( 4012): 	at com.amplitude.api.AmplitudeClient.isWithinMinTimeBetweenSessions(AmplitudeClient.java:546)
04-18 17:51:39.859 E/AndroidRuntime( 4012): 	at com.amplitude.api.AmplitudeClient.startNewSessionIfNeeded(AmplitudeClient.java:510)
04-18 17:51:39.859 E/AndroidRuntime( 4012): 	at com.amplitude.api.AmplitudeClient$4.run(AmplitudeClient.java:599)
04-18 17:51:39.859 E/AndroidRuntime( 4012): 	at android.os.Handler.handleCallback(Handler.java:739)
04-18 17:51:39.859 E/AndroidRuntime( 4012): 	at android.os.Handler.dispatchMessage(Handler.java:95)
04-18 17:51:39.859 E/AndroidRuntime( 4012): 	at android.os.Looper.loop(Looper.java:148)
04-18 17:51:39.859 E/AndroidRuntime( 4012): 	at android.os.HandlerThread.run(HandlerThread.java:61)
```

Note that the problem only occurs if ```enableForegroundTracking()``` is used.

[...]

PR #99, opened 2016-04-19
Stop using `instance` in AmplitudeClient and use `this` instead

djih:
Addresses this crash: https://github.com/amplitude/Amplitude-Android/issues/97

Error occurs because some of the AmplitudeClient methods are private, and PinnedAmplitudeClient.enableForegroundTracking uses the wrong instance

Thanks to @dmnugent80 for the fix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 
 * Add helper method `getSessionId` to expose the current sessionId value.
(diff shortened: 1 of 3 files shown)


djih:
Actually I think the better solution would be to stop using `instance`, and use `this` instead. This way we don't need to override the method in PinnedAmplitudeClient

paladin8:
lgtm

Issue #104, opened 2016-05-18
DeviceInfo.getMostRecentLocation crashes the app


ipereziriarte:
Hi,

I'm seeing inside the production logs a small amount of crashes.
Looks like **FORCE_STOP_PACKAGES permission is missing**

### StackTrace

```
Fatal Exception: java.lang.SecurityException: Permission Denial: forceStopPackage() from pid=21114, uid=10091 requires android.permission.FORCE_STOP_PACKAGES
       at android.os.Parcel.readException(Parcel.java:1425)
       at android.os.Parcel.readException(Parcel.java:1379)
       at android.location.ILocationManager$Stub$Proxy.getProviders(ILocationManager.java:851)
       at android.location.LocationManager.getProviders(LocationManager.java:321)
       at com.amplitude.api.DeviceInfo.getMostRecentLocation(DeviceInfo.java:309)
       at com.amplitude.api.DeviceInfo$CachedInfo.getCountryFromLocation(DeviceInfo.java:127)
       at com.amplitude.api.DeviceInfo$CachedInfo.getCountry(DeviceInfo.java:110)
       at com.amplitude.api.DeviceInfo$CachedInfo.(DeviceInfo.java)
       at com.amplitude.api.DeviceInfo$CachedInfo.(DeviceInfo.java)
       at com.amplitude.api.DeviceInfo.getCachedInfo(DeviceInfo.java:237)
       at com.amplitude.api.DeviceInfo.prefetch(DeviceInfo.java:243)
       at com.amplitude.api.AmplitudeClient$1.run(AmplitudeClient.java:163)
       at android.os.Handler.handleCallback(Handler.java:725)
       at android.os.Handler.dispatchMessage(Handler.java:92)
       at android.os.Looper.loop(Looper.java:153)
       at android.os.HandlerThread.run(HandlerThread.java:60)
```

### Environment

Amplitude version 2.7.1
Affected environment: production
Android version: Android 4.x and Android 5.x
Devices: 
+ huawei g510_0200
+ samsung_sgh_i527
+ samsung gt_i9060i
+ samsung sm_g530t1
+ vivo 4_8 hd

Thanks for your support.

djih:
@ipereziriarte does this crash only occur on those 5 devices? I wonder if it's a device-specific issue, like their manufacturer restricts access to GPS

djih:
[...]

PR #107, opened 2016-05-19
Catch security exceptions thrown during getProviders and location lookup

djih:
According to https://github.com/amplitude/Amplitude-Android/issues/104, it seems that trying to fetch the location information might sometimes fail on select devices. We want to catch the exception and make sure it doesn't crash the app/SDK.

Not sure why that security exception is occurring though, fetching location information shouldn't need anything other than location permissions. It seems to be isolated to select Android devices


paladin8:
lgtm

PR #108, opened 2016-06-22
Add support for Amazon advertising ID

jcomo:
Getting the advertiser ID currently only supports devices with Google play services. This adds support for getting the advertising ID from an Amazon device.

[Amazon advertising ID docs](https://developer.amazon.com/public/solutions/devices/fire-tablets/app-development/using-the-advertising-id-in-your-app)

### DIFF
--- a/src/com/amplitude/api/AmplitudeClient.java
+++ b/src/com/amplitude/api/AmplitudeClient.java
@@ -1699,6 +1699,7 @@ public class AmplitudeClient {
         invalidDeviceIds.add("000000000000000"); // Common Serial Number
         invalidDeviceIds.add("Android");
         invalidDeviceIds.add("DEFACE");
+        invalidDeviceIds.add("00000000-0000-0000-0000-000000000000");
 
         return invalidDeviceIds;
     }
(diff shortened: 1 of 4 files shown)


djih:
@jcomo thanks for the PR! We've tested it out on our end and it looks good. It looks like you're importing 2 new libraries `Secure` and `ContentResolver`, both of which have been available since Android API 3, so that should be fine. Thanks for adding a test too.

There is only 1 issue, which is we do not have an Amazon testing device, so we aren't able to verify with a real test case if it can successfully fetch the Amazon ADID. If you guys are able to verify that for us, then we should be good to release this as a new version.

jcomo:
Hey thanks for taking a look! We've tested it on two different versions of the Fire. Don't rush the release if you have other things to get in - we're good with running on a patched version for now

djih:
@jcomo this is merged and released in [v2.8.0](https://github.com/amplitude/Amplitude-Android/releases/tag/v2.8.0), which will become available on maven within a few hours. Thanks for the PR!

jcomo:
Thanks dude. We appreciate it 👍

PR #109, opened 2016-06-27
Init on background thread

djih:
* Run the `initialize` logic on the background thread so that the SQLite database operations do not delay the main thread.
* The `setUserId` and `setDeviceId` also have to be run on the background thread to maintain ordering


src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -1372,13 +1379,19 @@ public String getUserId() {
...
             return this;
         }
 
-        this.userId = userId;
-        dbHelper.insertOrReplaceKeyValue(USER_ID_KEY, userId);
+        final AmplitudeClient client = this;
+        runOnLogThread(new Runnable() {
+            @Override
+            public void run() {
+                client.userId = userId;
```
curtisliu: should this be done outside of the run on log thread?

djih: It's done on the log thread to maintain ordering, say if you already have 2 log events queued up when you set a new user id, you want those 2 events to still get the old user id.

curtisliu:
Lgtm

PR #110, opened 2016-07-05
Flush unsent events on app close or minimize

djih:
This adds a call to `updateServer()` in the `onExitForeground` handler, called by `onActivityPaused` activity lifecycle callback, which will flush events when the app goes into the background / closed / minimized (matching the behavior of our iOS SDK). This only works assuming the user installs the callbacks, which is recommended by default in the README. Verified with unit test and in demo app that going to background forces the flush.

Some considerations: depending on how strict Android is, they may or may not allow the flush to complete fully if the app is being killed. In update server we only upload in batches of max 30 events, so there's a good chance we can at least get a single batch out before being killed. The important thing is that the uploading + updating of database happens together. If the process is killed before the database is updated, then there is the potential for duplicate uploading of events. 

I think it should be okay though, Android seems to be pretty generous with the Activity Lifecycle: https://developer.android.com/training/basics/activity-lifecycle/stopping.html. In this article they recommend putting database updating logic in the onStop callback, so it seems like they allow those to completion. We're putting it in the onPause, since we don't know if the user is putting the app to the background and returning, or if they're putting the app to the background and just forgetting about the app. In both cases we still want to flush, and those would need to be handled by onPause.


curtisliu:
The one concern I have here is that this will force flushing of events much more frequently if the app is composed of multiple activities. I think it should be an option instead of always on

djih:
As per our discussion, leaving the flushing as default behavior, adding a `setFlushEventsOnClose(boolean)` option to disable. Updated CHANGELOG, README and documentation files. PTAL @curtisliu

curtisliu:
LGTM

PR #111, opened 2016-07-06
Update revenue

djih:
* productId is no longer a required field. This is required to update Segment Android SDK to use logRevenueV2 since it assumes productId can be null
* Do not truncate receipt and receipt sig fields even if they exceed 1024 characters
* regenerated docs to update the productId comments


curtisliu:
LGTM

PR #112, opened 2016-07-08
migrate database file

djih:
Added 1 time renaming of sqlite database file during the initialization of the DatabaseHelper instance.

This was a little tricky to unit test. I ended up writing a separate SqliteDatabaseHelper class to re-create the old database file, and then forcing the migration path. I tested it with the demo app in the simulator and in the debugger walked through the migration to make sure it successfully renamed and once more to make sure it skipped after the initial migration.


djih:
Actually we can simplify the testing by allowing null apiKeys in `getDatabaseHelper` to perform the old behavior, that way we can easily re-create the old database file, add test values, and test the migration

CHANGELOG.md
```diff
@@ -10,6 +12,7 @@
 * Fix bug where `setOptOut` was not being run on background thread.
 * `productId` is no longer a required field for `Revenue` logged via `logRevenueV2`.
 * Fix bug where receipt and receiptSignature were being truncated if they were too long (exceeded 1024 characters).
+>>>>>>> master
```
djih: oops will clean this up before pushing

curtisliu:
lgtm

haoliu-amp:
@djih I think this PR is not needed, right?

djih:
@haoliu-amp yea I think I already merged similar code a long time ago

Issue #113, opened 2016-07-11
AmplitudeClient#setOptOut() blocks the main thread


austynmahoney:
[`AmplitudeClient#setOptOut()`](https://github.com/amplitude/Amplitude-Android/blob/master/src/com/amplitude/api/AmplitudeClient.java#L426) runs on the main thread.

I call this method directly after `initialize()`, so it looks like the database has not been created yet due to a race condition from the backgrounded `initialize()` call. This can cause the main thread to be blocked for around 150ms on start. On subsequent calls (once the database has been created), it can still take around 70ms on an emulator.

Any method that has database calls should probably be put into some kind of background task queue. That way they are executed in order, and do not block the main thread.

```
StrictMode policy violation; ~duration=150 ms: android.os.StrictMode$StrictModeDiskReadViolation: policy=65567 violation=2
                  at android.os.StrictMode$AndroidBlockGuardPolicy.onReadFromDisk(StrictMode.java:1263)
                  at libcore.io.BlockGuardOs.stat(BlockGuardOs.java:292)
                  at java.io.File.isDirectory(File.java:522)
                  at android.app.ContextImpl.validateFilePath(ContextImpl.java:1931)
                  at android.app.ContextImpl.openOrCreateDatabase(ContextImpl.java:566)
                  at android.content.ContextWrapper.openOrCreateDatabase(ContextWrapper.java:269)
                  at android.database.sqlite.SQLiteOpenHelper.getDatabaseLocked(SQLiteOpenHelper.java:223)
                  at android.database.sqlite.SQLiteOpenHelper.getWritableDatabase(SQLiteOpenHelper.java:163)
                  at com.amplitude.api.DatabaseHelper.insertOrReplaceKeyValueToTable(DatabaseHelper.java:127)
                  at com.amplitude.api.DatabaseHelper.insertOrReplaceKeyLongValue(DatabaseHelper.java:121)
                  at com.amplitude.api.AmplitudeClient.setOptOut(AmplitudeClient.java:432)
```

djih:
@austynmahoney Thanks for reporting, this was supposed to have also been updated with the latest version. Will patch and release a new version

djih:
@austynmahoney thanks again for reporting. The fix has been pushed with [v2.9.1](https://github.com/amplitude/Amplitude-Android/releases/tag/v2.9.1), it will show up in Maven within a few hours

austynmahoney:
@djih Thanks!

[...]

PR #114, opened 2016-07-11
setOptOut on background thread

djih:
Addresses https://github.com/amplitude/Amplitude-Android/issues/113
With this [change](https://github.com/amplitude/Amplitude-Android/pull/109/files) that went out in v2.8.0, I forgot to also update `setOptOut` to run on background thread.


curtisliu:
LGTM

PR #115, opened 2016-07-12
revenue equals and hashcode

djih:


djih:
Planning to push this as v2.9.2 to support Segment integration unit tests

curtisliu:
LGTM, is it possible to make sure the segment integration unit tests work with this before pushing?

djih:
Good call with the testing. Turns out since eventProperties is a JSONObject, you can't do a straight comparison in the equals method. I refactored the `compareJSONObjects` method that we already had for testing into the `Utils` class. I verified that the Segment unit tests now pass. PTAL @curtisliu

djih:
Also cleaned up documentation updates, we will just regenerate all the docs right before releasing a new version to avoid polluting the PR diffs

PR #117, opened 2016-07-14
set location listening on background thread

elevenfive:
Otherwise, can't set location listening reliably -- likely to throw exception (see below), since init is on background thread, and no way to get callback when init completed, in order to enable/disable location listening.

```
java.lang.IllegalStateException: Must initialize before acting on location listening.
at com.amplitude.api.AmplitudeClient.disableLocationListening(AmplitudeClient.java:336)
```


djih:
@elevenfive thanks for the PR, merging to master and pushing new version now

djih:
@elevenfive released [v2.9.2](https://github.com/amplitude/Amplitude-Android/releases/tag/v2.9.2), should show up in Maven central within an hour or so

Issue #42, opened 2015-05-14
Still causing app crashes due to failed SQL allocation


xiphirx:
Since it seems like issue #19 is not being looked at, the issue still persists even after updating:

Still exists on v1.6.2,

Device types:
![image](https://cloud.githubusercontent.com/assets/993832/7528014/354dbb96-f4d7-11e4-840f-29e61f967872.png)

Android versions:
![image](https://cloud.githubusercontent.com/assets/993832/7528016/3f519450-f4d7-11e4-9375-778c90af93ed.png)

```
0	android.database.CursorWindowAllocationException: Cursor window allocation of 2048 kb failed.
1	at android.database.CursorWindow.<init>(CursorWindow.java:109)
2	at android.database.CursorWindow.<init>(CursorWindow.java:100)
3	at android.database.AbstractWindowedCursor.clearOrCreateWindow(AbstractWindowedCursor.java:198)
4	at android.database.sqlite.SQLiteCursor.clearOrCreateWindow(SQLiteCursor.java:301)
5	at android.database.sqlite.SQLiteCursor.fillWindow(SQLiteCursor.java:139)
6	at android.database.sqlite.SQLiteCursor.getCount(SQLiteCursor.java:133)
7	at android.database.AbstractCursor.moveToPosition(AbstractCursor.java:197)
8	at android.database.AbstractCursor.moveToNext(AbstractCursor.java:245)
9	at com.amplitude.api.DatabaseHelper.getEvents(DatabaseHelper.java:91)
10	at com.amplitude.api.AmplitudeClient.updateServer(AmplitudeClient.java:594)
11	at com.amplitude.api.AmplitudeClient.updateServer(AmplitudeClient.java:582)
12	at com.amplitude.api.AmplitudeClient.access$1400(AmplitudeClient.java:31)
13	at com.amplitude.api.AmplitudeClient$4.run(AmplitudeClient.java:343)
14	at android.os.Handler.handleCallback(Handler.java:733)
15	at android.os.Handler.dispatchMessage(Handler.java:95)
16	at android.os.Looper.loop(Looper.java:136)
17	at android.os.HandlerThread.run(HandlerThread.java:61)
```

curtisliu:
Yes, we're aware of this and still looking into the issue. Will keep you updated.

niqo01:
@xiphirx just curious, How often this issue occur?

benjamin-bader:
[...]

PR #123, opened 2016-10-07
Gracefully catch and handle CursorWindowAllocationException

djih:
This addresses: https://github.com/amplitude/Amplitude-Android/issues/42. We believe that our SDK correctly closes cursors, which would mean that these CursorWindowAllocationExceptions are caused by the app running out of memory, not something in our SDK's control. That being said, the crash still originates from our SDK, and so we should attempt to catch it and gracefully handle it, instead of crashing the app.

Our SDK uses cursors when fetching unsent events/identifies and key/value pairs. We need to catch the exception in these 3 primary use cases:

1. Fetching unsent events/identifies for upload
    * If the fetch fails, then we can simply defer the upload, and try again later maybe when the app has more free memory
2. Fetching the userId/deviceId during initialization
    * It is crucial that our SDK loads the existing userId/deviceIds during initialization, otherwise the tracked events will go to the wrong user and affect new and active user counts. If we cannot load those ids, then we cannot log any events.
    * We can set the apiKey back to null in the catch handler, as if `initialize` was never called on the SDK. All of the core SDK methods already check that an apiKey is set, and so they will just be ignored
    * The key detail here is the initialization logic is run on the background thread, so it's possible that the apiKey is set to null after a bunch of logEvents have already been scheduled (passed the apiKey check), so then we need to add an additional apiKey check inside all of the runnables
3. Fetching event metadata during logEvent
    * Whenever the SDK logs an event, it fetches some meta data like the last sequence number, increments it, sets it on the event, and then saves the new value back to the database.
    * We don't want the meta data fetching to fail, otherwise the events could end up in a weird state. What we should do is load all of the meta data into memory during initialize, and then just increment and read the values in memory. We still need to write any value updates to the database, but those don't require a Cursor object, so those should not fail

Couple extra things to note:
* I was not able to directly import the `CursorWindowAllocationException` class. I can see it in the Android Source, but I'm not able to access it directly. So instead what I have to do is catch the base class `RuntimeException` and do a string check on the message. If the exception caught is not a CursorWindowAllocationException, we need to rethrow it
* Since the entire initialization logic was moved to a background thread, the `initializeDeviceInfo` does not need to be explicitly wrapped in the `runOnLogThread`. That wrapping also prevents unit tests from running correctly since the logThread is improperly named to "main" thread in unit testing

### DIFF
--- a/src/com/amplitude/api/AmplitudeClient.java
+++ b/src/com/amplitude/api/AmplitudeClient.java
@@ -13,7 +13,6 @@ import com.amplitude.security.MD5;
 import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
-import org.w3c.dom.Text;
 
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
@@ -452,6 +451,10 @@ public class AmplitudeClient {
         runOnLogThread(new Runnable() {
             @Override
             public void run() {
+                // one more apiKey check, in case initialization failed
+                if (TextUtils.isEmpty(client.apiKey)) {
+                    return;
+                }
                 client.optOut = optOut;
                 dbHelper.insertOrReplaceKeyLongValue(OPT_OUT_KEY, optOut ? 1L : 0L);
             }
@@ -697,8 +700,8 @@ public class AmplitudeClient {
      * @see <a href="https://github.com/amplitude/Amplitude-Android#setting-groups">
      *     Setting Groups</a>
      */
-    public void logEventSync(String eventType, JSONObject eventProperties, JSONObject group) {
-        logEventSync(eventType, eventProperties, group, false);
+    public void logEventSync(String eventType, JSONObject eventProperties, JSONObject groups) {
+        logEventSync(eventType, eventProperties, groups, false);
     }
 
     /**
@@ -777,9 +780,14 @@ public class AmplitudeClient {
         final JSONObject copyEventProperties = eventProperties;
         final JSONObject copyUserProperties = userProperties;
         final JSONObject copyGroups = groups;
+        final AmplitudeClient client = this;
         runOnLogThread(new Runnable() {
             @Override
             public void run() {
+                // one more apiKey check, in case initialization failed
+                if (TextUtils.isEmpty(client.apiKey)) {
+                    return;
+                }
                 logEvent(
                     eventType, copyEventProperties, apiProperties,
                     copyUserProperties, copyGroups, timestamp, outOfSession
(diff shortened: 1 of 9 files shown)


src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -916,63 +945,38 @@ private long getLongvalue(String key, long defaultValue) {
...
         dbHelper.insertOrReplaceKeyLongValue(SEQUENCE_NUMBER_KEY, sequenceNumber);
         return sequenceNumber;
     }
 
     /**
-     * Internal method to get the last event time.
-     *
-     * @return the last event time
-     */
-    long getLastEventTime() {
```
curtisliu: My sense is we can leave these methods and just change the implementation

src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -1542,6 +1554,19 @@ public void run() {
             } catch (JSONException e) {
                 uploadingCurrently.set(false);
                 logger.e(TAG, e.toString());
+
+            // handle CursorWindowAllocationException when fetching events, defer upload
+            } catch (RuntimeException e) {
+                String message = e.getMessage();
+                if (!TextUtils.isEmpty(message) && message.startsWith("Cursor window")) {
```
curtisliu: Can this logic be factored into a method?

src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -225,28 +234,48 @@ public synchronized AmplitudeClient initialize(final Context context, final Stri
...
-                        client.userId = userId;
-                        client.dbHelper.insertOrReplaceKeyValue(USER_ID_KEY, userId);
-                    } else {
-                        client.userId = client.dbHelper.getValue(USER_ID_KEY);
-                    }
-                    Long optOut = client.dbHelper.getLongValue(OPT_OUT_KEY);
-                    client.optOut = optOut != null && optOut == 1;
+                if (!initialized) {
+                    try {
+                        AmplitudeClient.upgradePrefs(context);
```
curtisliu: verify that all steps in this try/catch block are idempotent and add a comment that they should be

src/com/amplitude/api/AmplitudeClient.java
djih: Refactored the message checking part into a method

src/com/amplitude/api/AmplitudeClient.java
djih: I feel like it's cleaner if we just remove them? The thing is like the calling methods will save the values to local variables that have the same name as the instance variables. Having duplicate variables + extra level of indirection is kind of messy

curtisliu: Yeah, I agree with you.

src/com/amplitude/api/AmplitudeClient.java
djih: verified and commented

djih:
Pulled the exception handling logic into the DatabaseHelper. The DatabaseHelper will throw an Amplitude `cursorWindowAllocationException` which can then be caught cleanly by the SDK client. I had to wrap db.query to facilitate mocking during unit tests to throw exceptions.

You can take another look if you want @curtisliu

PR #124, opened 2016-10-11
Diagnostics

djih:
Added a stand-alone `Diagnostics` class that handles the logging and uploading of error events. Error events are queued up in a new database table, which bumps up the DB version to 4. The `Diagnostics` class has its own background thread so that it doesn't block the SDK's logThread and httpThreads, but it still re-uses the AmplitudeClient's OkHttpClient, and uploads events to a new dedicated endpoint. Error events are uploaded when the SDK uploads events / identifies. The error events themselves are very lightweight, containing only 2 fields: an `error` string, and a `timestamp`.

Currently it only logs an error when the SDK fails to upload a `start_session` event, and records the type of error (for example no_connection, 413, etc).

New public methods:
- `enableDiagnosticLogging(true)`: this enables the logging of upload failure errors. Call this after initializing the AmplitudeClient with an apiKey
- `setDiagnosticEventMaxCount(50)`: similar to events, we queue up unsent error events in a database table. This limits the size and memory footprint. By default after 50 unsent error events it will start dropping events starting with the earliest
- `enableFlushSessionEventsOnLog(true)`: this forces the SDK to flush start/end session events immediately when it starts a new session.


src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -1708,26 +1764,39 @@ public void run() {
             } else {
                 logger.w(TAG, "Upload failed, " + stringResponse
                         + ", will attempt to reupload later");
+                diagnosticError = "unknown";
```
curtisliu: make this the response code

src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -320,6 +322,50 @@ public AmplitudeClient enableNewDeviceIdPerInstall(boolean newDeviceIdPerInstall
...
+     * @return the AmplitudeClient
+     */
+    public AmplitudeClient enableDiagnosticLogging(boolean enableDiagnosticLogging) {
+        if (!contextAndApiKeySet("enableDiagnosticLogging")) {
+            return this;
+        }
+
+        this.logDiagnosticEvents = enableDiagnosticLogging;
+        if (enableDiagnosticLogging) {
+            Diagnostics.getLogger(context).enableLogging(httpClient, apiKey);
```
djih: this actually needs to be run on the logThread to make sure initialize finishes running, otherwise httpClient will be null even though there might be an apikey

PR #125, opened 2016-10-26
identify with outOfSession

djih:
Some customers want to log identify events while their app is in the background, without triggering start/end session events. We can expose the outOfSession boolean as an argument.

### DIFF
--- a/src/com/amplitude/api/AmplitudeClient.java
+++ b/src/com/amplitude/api/AmplitudeClient.java
@@ -1297,12 +1297,27 @@ public class AmplitudeClient {
      *     User Properties</a>
      */
     public void identify(Identify identify) {
+        identify(identify, false);
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -1297,12 +1297,29 @@ public void clearUserProperties() {
...
+     * @param outOfSession whther to log the identify event out of session
+     * @see com.amplitude.api.Identify
+     * @see <a href="https://github.com/amplitude/Amplitude-Android#user-properties-and-user-property-operations">
+     *     User Properties</a>
+     */
+    public void identify(Identify identify, boolean outOfSession) {
+        if (
+            identify == null || identify.userPropertiesOperations.length() == 0 ||
+            !contextAndApiKeySet("identify()")
+        ) return;
```
paladin8: This feels weird stylistically to me, but w/e.

PR #129, opened 2016-12-22
Handle null event strings

djih:
Android SDK crashes when loading events from db table if there's a null event string.

* check event string is not null before deserializing - this should stop the crashes
* it's important to note that the metadata state is still okay (like if it logged a null event, then lastEventId will be incremented, but it can safely skip over the null event when fetching)
* if valid events are logged after the null event, they still get sent correctly
* event servers can handle payloads with empty event lists (although if there are no events to send the SDK won't even make a request out)
* Added some more logging around JSON serialization errors - I think that's how null events are getting added to the table
* Only save the event string if it serializes properly

### DIFF
--- a/test/com/amplitude/api/DatabaseHelperTest.java
+++ b/test/com/amplitude/api/DatabaseHelperTest.java
@@ -484,7 +484,7 @@ public class DatabaseHelperTest extends BaseTest {
     }
 
     @Test
-    public void testNullEventSTring() throws JSONException {
+    public void testNullEventString() throws JSONException {
         dbInstance.addEvent(null);
         List<JSONObject> events = dbInstance.getEvents(-1, -1);
         assertTrue(events.isEmpty());
--- a/test/com/amplitude/api/DatabaseHelperTest.java
+++ b/test/com/amplitude/api/DatabaseHelperTest.java
@@ -14,6 +14,7 @@ import java.util.List;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 @RunWith(RobolectricTestRunner.class)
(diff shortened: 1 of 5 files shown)


djih:
Stack trace of crash:
```java
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.String.length()' on a null object reference
       at org.json.JSONTokener.nextCleanInternal(JSONTokener.java:116)
       at org.json.JSONTokener.nextValue(JSONTokener.java:94)
       at org.json.JSONObject.(JSONObject.java:156)
       at org.json.JSONObject.(JSONObject.java:173)
       at com.amplitude.api.DatabaseHelper.getEventsFromTable$20f06a58(Unknown Source:253)
       at com.amplitude.api.DatabaseHelper.getEvents(Unknown Source:231)
       at com.amplitude.api.AmplitudeClient.updateServer(Unknown Source:1522)
       at com.amplitude.api.AmplitudeClient.updateServer(Unknown Source:1494)
       at com.amplitude.api.AmplitudeClient$7.run(Unknown Source:1117)
       at android.os.Handler.handleCallback(Handler.java:739)
       at android.os.Handler.dispatchMessage(Handler.java:95)
       at android.os.Looper.loop(Looper.java:135)
       at android.os.HandlerThread.run(HandlerThread.java:61)
```

test/com/amplitude/api/DatabaseHelperTest.java
```diff
@@ -481,5 +482,12 @@ public void testNoConflictBetweenEventsAndIdentifys() {
         assertEquals(0, dbInstance.getEventCount());
         assertEquals(1, dbInstance.getIdentifyCount());
     }
+
+    @Test
+    public void testNullEventSTring() throws JSONException {
```
paladin8: String

djih:
@paladin8 notes that the `JSONObject.toString()` can return null when it catches an exception: http://www.docjar.com/html/api/org/json/JSONObject.java.html (line 1341)

PR #132, opened 2017-02-15
Added support for multiple apps (separate API keys) using non-singleton approach

gefilte:
**Problem**: we maintain an Android library that we share between apps in our company that we wish to produce its own tracking with Amplitude SDK. The problem is that because Amplitude SDK only works with one app at a time, if an app using this library *also* uses Amplitude SDK tracking, as soon as our library code is started it will "take over" the logging of events, stopping the flow of events for the hosting app's API key.

**Solution**
With this change, the singleton approach to Amplitude SDK should work completely unchanged – this change is backwards-compatible. In order to have multiple Amplitude apps (API keys) in a single app, you can use the public constructor for AmplitudeClient and retain a reference to it in wrapper code in your app and/or library.

For example:

```
amplitudeClient = new AmplitudeClient();
amplitudeClient.initialize(app, apiKey).enableForegroundTracking(app);
```
...later...
```
amplitudeClient.setUserId(realmId);
amplitudeClient.logEvent(eventName, new JSONObject(properties));
```

The AmplitudeClient class was already written in such a way as not to depend on being a singleton: the only heavy lifting that was necessary in this pull request was to have each client depend on a separate  SQLite DB. (Otherwise events would "bleed" across instances.)

It would also be easy to make the ```getInstance()``` accept a key string as well, though using the apiKey as the DB name key seemed a sensible way to handle this. (It will prevent you from instantiating the same app twice.)

Please consider this for integration, it solves a real problem we are facing now.

Thanks!

### DIFF
--- a/src/com/amplitude/api/AmplitudeClient.java
+++ b/src/com/amplitude/api/AmplitudeClient.java
@@ -228,7 +228,7 @@ public class AmplitudeClient {
 
         this.context = context.getApplicationContext();
         this.apiKey = apiKey;
-        this.dbHelper = DatabaseHelper.getDatabaseHelper(this.context);
+        this.dbHelper = DatabaseHelper.getDatabaseHelper(this.context, apiKey);
 
         final AmplitudeClient client = this;
         runOnLogThread(new Runnable() {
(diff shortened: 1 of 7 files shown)


djih:
@gefilte thanks for the PR! we already have a long standing branch for this: https://github.com/amplitude/Amplitude-Android/pull/85, although it's been deprioritized since we haven't had any requests for it.

What is your use case for this? I can pass the feedback to our product team

gefilte:
I detailed the use case above. In summary, we have a library we share between apps that we wish to send tracking events to its own Amplitude "app", so we can do analysis of how that library is used across apps. Those apps however need their own tracking, which should be inclusive of the events coming from the library.

djih:
@gefilte sorry for the delay but we've officially added support for logging events to multiple instances with v2.14.0: https://github.com/amplitude/Amplitude-Android/releases/tag/v2.14.0. It should be available on Maven within a few hours

PR #141, opened 2017-04-27
Catch more exceptions from TelephonyManager

fkam-tt:
I've come across exceptions in `getCarrier` when initializing the Amplitude client. This change handles this in the same way as `getCountryFromNetwork`.


djih:
Thanks @fkam-tt this patch will go out in our next version which will go out later today

djih:
@fkam-tt this patch has been released on [v2.13.4](https://github.com/amplitude/Amplitude-Android/releases/tag/v2.13.4) which should be available on Maven in an hour or so

PR #144, opened 2017-07-05
add support for multiple instances

djih:
Feature parity with our iOS / JS SDKs.

Instead of 1 static instance of the AmplitudeClient, we have the Amplitude parent class maintain a map from instance name to AmplitudeClient. 

New calling convention: `Amplitude.getInstance("app name").<function>`
Existing calls to `Amplitude.getInstance()` will get mapped to the default instance (we catch null instance names and substitute with "$default_instance"`.
Instance names are case insensitive by design choice

Need to add an instance map in 3 places: Amplitude parent class, DatabaseHelper class (each instance needs its own separate database file, the default instance keeps the existing database file so we don't need any migration code), PinnedAmplitudeClient

### DIFF
--- a/src/com/amplitude/api/Amplitude.java
+++ b/src/com/amplitude/api/Amplitude.java
@@ -1,7 +1,6 @@
 package com.amplitude.api;
 
 import android.content.Context;
-import android.text.TextUtils;
 
 import org.json.JSONObject;
 
@@ -38,12 +37,8 @@ public class Amplitude {
      * @param instance name to get "ex app 1"
      * @return the specified instance
      */
-    public static synchronized  AmplitudeClient getInstance(String instance) {
-        if (TextUtils.isEmpty(instance)) {
-            instance = Constants.DEFAULT_INSTANCE;
-        }
-        instance = instance.toLowerCase();
-
+    public static synchronized AmplitudeClient getInstance(String instance) {
+        instance = Utils.normalizeInstanceName(instance);
         AmplitudeClient client = instances.get(instance);
         if (client == null) {
             client = new AmplitudeClient(instance);
--- a/src/com/amplitude/api/Amplitude.java
+++ b/src/com/amplitude/api/Amplitude.java
@@ -1,8 +1,12 @@
 package com.amplitude.api;
 
 import android.content.Context;
+import android.text.TextUtils;
 
 import org.json.JSONObject;
+
+import java.util.HashMap;
+import java.util.Map;
 
 
 /**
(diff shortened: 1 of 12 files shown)


src/com/amplitude/api/Amplitude.java
```diff
@@ -16,14 +20,36 @@
...
+    }
+
+    /**
+     * Gets the specified instance. If instance is null or empty string, fetches the default
+     * instance instead.
+     *
+     * @param instance name to get "ex app 1"
+     * @return the specified instance
+     */
+    public static synchronized  AmplitudeClient getInstance(String instance) {
```
paladin8: Extra space after `synchronized`.

src/com/amplitude/api/DatabaseHelper.java
```diff
@@ -47,19 +49,43 @@
...
-        if (instance == null) {
-            instance = new DatabaseHelper(context.getApplicationContext());
+        return getDatabaseHelper(context, null);
+    }
+
+    static synchronized DatabaseHelper getDatabaseHelper(Context context, String instance) {
+        if (TextUtils.isEmpty(instance)) {
+            instance = Constants.DEFAULT_INSTANCE;
+        }
+        instance = instance.toLowerCase();
```
paladin8: Maybe these few lines of logic should be refactored into a `normalizeInstanceName` method?

src/com/amplitude/api/DatabaseHelper.java
```diff
@@ -47,19 +49,43 @@
...
+
+        DatabaseHelper dbHelper = instances.get(instance);
+        if (dbHelper == null) {
+            if (instance.equalsIgnoreCase(Constants.DEFAULT_INSTANCE)) {
```
paladin8: Feels like the `DatabaseHelper` constructor may as well do this check instead of it being outside.

src/com/amplitude/api/Amplitude.java
djih: good catch

src/com/amplitude/api/DatabaseHelper.java
djih: good idea

src/com/amplitude/api/DatabaseHelper.java
djih: I think you're required to have the default constructor (without the instance arg). Also can't have any branching in the constructor logic, the first line needs to be `super(`

PR #161, opened 2018-04-19
Set user id new session

djih:
Expose a public `setUserIdAndStartNewSession` method that is basically https://github.com/amplitude/Amplitude-Android/blob/master/src/com/amplitude/api/AmplitudeClient.java#L1108-L1120 and sets the userId between sending the start and end session events.

### DIFF
--- a/test/com/amplitude/api/SessionTest.java
+++ b/test/com/amplitude/api/SessionTest.java
@@ -1048,4 +1048,83 @@ public class SessionTest extends BaseTest {
         // the new event should have a newer session id
         assertTrue(session_event.optLong("session_id") > timestamp);
     }
+
+    @Test
+    public void testSetUserIdAndDoNotStartNewSessionWithTracking() {
+        amplitude.trackSessionEvents(true);
+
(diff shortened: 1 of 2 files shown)


test/com/amplitude/api/SessionTest.java
```diff
@@ -992,4 +992,60 @@ public void testOutOfSessionIdentifyDoesNotTriggerNewSession() throws JSONExcept
...
+
+        long timestamp = System.currentTimeMillis();
+        amplitude.logEventAsync("test", null, null, null, null, timestamp, false);
+        Shadows.shadowOf(amplitude.logThread.getLooper()).runToEndOfTasks();
+
+        // trackSessions is true, start_session event is added
+        assertEquals(getUnsentEventCount(), 2);
+
+        // set user id and validate session ended and new session started
+        amplitude.setUserId("test_new_user", true);
```
paladin8: Maybe test a call with false to make sure there are no extra events being generated? And try a call with session events off?

djih: good idea. done

PR #170, opened 2018-07-20
Add tracking config

djih:
Add ability to block tracking of specific properties in Android SDK. Created a helper class `TrackingConfig` that tracks fields to disable in a dict, and the exposes package private `shouldTrack<Field>` methods. We also want to send a subset of these disabled properties in the `api_properties` field server-side so that we can block those properties server-side.


src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -908,33 +921,61 @@ protected long logEvent(String eventType, JSONObject eventProperties, JSONObject
...
+            if (trackingOptions.shouldTrackOsVersion()) {
+                event.put("os_version", replaceWithJSONNull(deviceInfo.getOsVersion()));
+            }
+            if (trackingOptions.shouldTrackDeviceBrand()) {
+                event.put("device_brand", replaceWithJSONNull(deviceInfo.getBrand()));
+            }
+            if (trackingOptions.shouldTrackDeviceManufacturer()) {
+                event.put("device_manufacturer", replaceWithJSONNull(deviceInfo.getManufacturer()));
+            }
+            if (trackingOptions.shouldTrackDeviceModel()) {
```
djih: Unclear how much this affects performance, basically for each of these fields we need to do a Set contains check. For each field, for each event.

src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -908,33 +921,61 @@ protected long logEvent(String eventType, JSONObject eventProperties, JSONObject
...
             event.put("library", library);
 
             apiProperties = (apiProperties == null) ? new JSONObject() : apiProperties;
-            Location location = deviceInfo.getMostRecentLocation();
-            if (location != null) {
-                JSONObject locationJSON = new JSONObject();
-                locationJSON.put("lat", location.getLatitude());
-                locationJSON.put("lng", location.getLongitude());
-                apiProperties.put("location", locationJSON);
+            JSONObject apiPropertiesTrackingOptionsCopy = Utils.cloneJSONObject(apiPropertiesTrackingOptions);
```
djih: api properties was already cloned before calling `logEvent`, so should be safe to mutate.
We don't want to mutate the `apiPropertiesTrackingOptions` blob in memory since that gets copied to each event.

src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -908,33 +921,60 @@ protected long logEvent(String eventType, JSONObject eventProperties, JSONObject
...
 
             apiProperties = (apiProperties == null) ? new JSONObject() : apiProperties;
-            Location location = deviceInfo.getMostRecentLocation();
-            if (location != null) {
-                JSONObject locationJSON = new JSONObject();
-                locationJSON.put("lat", location.getLatitude());
-                locationJSON.put("lng", location.getLongitude());
-                apiProperties.put("location", locationJSON);
+            if (apiPropertiesTrackingOptions.length() > 0) {
+                apiProperties.put("tracking_options", apiPropertiesTrackingOptions);
```
djih: api properties was already cloned before calling `logEvent`, so should be safe to mutate

src/com/amplitude/api/TrackingOptions.java
```diff
@@ -0,0 +1,193 @@
...
+        Constants.AMP_TRACKING_OPTION_COUNTRY,
+        Constants.AMP_TRACKING_OPTION_DMA,
+        Constants.AMP_TRACKING_OPTION_IP_ADDRESS,
+        Constants.AMP_TRACKING_OPTION_LAT_LON,
+        Constants.AMP_TRACKING_OPTION_REGION,
+    };
+
+    public static final String TAG = "com.amplitude.api.TrackingOptions";
+
+    Set<String> disabledFields = new HashSet<String>();
```
paladin8: Maybe these should be enums instead of strings? I don't think we use the string form ever, but it's not that big of a deal either way.

src/com/amplitude/api/TrackingOptions.java
```diff
@@ -0,0 +1,193 @@
...
+        return shouldTrackField(Constants.AMP_TRACKING_OPTION_VERSION_NAME);
+    }
+
+    private void disableTrackingField(String field) {
+        disabledFields.add(field);
+    }
+
+    protected JSONObject getApiPropertiesTrackingOptions() {
+        JSONObject apiPropertiesTrackingOptions = new JSONObject();
+        if (disabledFields.size() == 0) {
```
paladin8: Prefer `isEmpty()`.

src/com/amplitude/api/TrackingOptions.java
```diff
@@ -0,0 +1,193 @@
...
+    private static String[] SERVER_SIDE_PROPERTIES = {
+        Constants.AMP_TRACKING_OPTION_CITY,
+        Constants.AMP_TRACKING_OPTION_COUNTRY,
+        Constants.AMP_TRACKING_OPTION_DMA,
+        Constants.AMP_TRACKING_OPTION_IP_ADDRESS,
+        Constants.AMP_TRACKING_OPTION_LAT_LON,
+        Constants.AMP_TRACKING_OPTION_REGION,
+    };
+
+    public static final String TAG = "com.amplitude.api.TrackingOptions";
```
paladin8: Seems like we want `TrackingOptions.class.getCanonicalName()`.

PR #173, opened 2018-08-08
Insert metadata on db reset

djih:
When encountering SQLite exception, our database helper discards the database and starts over. This discards important metadata such as device id and user id. We still have these in memory though, so when that happens we have a chance to re-insert into the fresh database before the app closes.

1. Added a listener that gets called on database reset to re-insert values
2. Just for good measure, in the onExitForeground, re-insert metadata one last time

Note: found it difficult to mock out the SqliteDatabaseHelper class during testing, especially around the `close()` method and `getWriteableDatabase()`. I ended up creating a helper method in insertEvent for the sole purpose of enabling mocking.


src/com/amplitude/api/DatabaseHelper.java
```diff
@@ -226,6 +232,10 @@ private synchronized long addEventToTable(String table, String event) {
         return result;
     }
 
+    synchronized long insertEventContentValuesIntoTable(SQLiteDatabase db, String table, ContentValues contentValues) throws SQLiteException, StackOverflowError {
```
djih: Adding a helper method / level of indirection here to support mocking

src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -284,6 +284,18 @@ public void run() {
...
+                        // install database reset listener to re-insert metadata in memory
+                        dbHelper.setDatabaseResetListener(new DatabaseResetListener() {
+                            @Override
+                            public void onDatabaseReset() {
+                                dbHelper.insertOrReplaceKeyValue(DEVICE_ID_KEY, deviceId);
+                                dbHelper.insertOrReplaceKeyValue(USER_ID_KEY, client.userId);
+                                dbHelper.insertOrReplaceKeyLongValue(OPT_OUT_KEY, optOut ? 1L : 0L);
+                                dbHelper.insertOrReplaceKeyLongValue(PREVIOUS_SESSION_ID_KEY, sessionId);
+                                dbHelper.insertOrReplaceKeyLongValue(LAST_EVENT_TIME_KEY, lastEventTime);
+                            }
```
curtisliu: Should we be persisting everything that we have from getLongvalue()? (sequenceNumber, lastEventId, lastIdentifyId, lastEventTime)

src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -1223,6 +1235,12 @@ public void run() {
                 if (flushEventsOnClose) {
                     updateServer();
                 }
+
+                // re-persist metadata into database for good measure
+                dbHelper.insertOrReplaceKeyValue(DEVICE_ID_KEY, deviceId);
+                dbHelper.insertOrReplaceKeyValue(USER_ID_KEY, userId);
+                dbHelper.insertOrReplaceKeyLongValue(OPT_OUT_KEY, optOut ? 1L : 0L);
+                dbHelper.insertOrReplaceKeyLongValue(LAST_EVENT_TIME_KEY, lastEventTime);
```
curtisliu: Persist sessionId to PREVIOUS_SESSION_ID_KEY and the other ones I mentioned above?

src/com/amplitude/api/AmplitudeClient.java
djih: sequence number, last event id, last identify id correspond to the state of the event / identify tables in the database. on db reset those tables are dropped, so I think those counters should be cleared

src/com/amplitude/api/AmplitudeClient.java
djih: persisting sessionId. The last event time gets refreshed a few lines above in the same function `refreshSessionTime`.

PR #174, opened 2018-10-10
add support for group identifies

djih:


src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -1025,7 +1034,7 @@ protected long saveEvent(String eventType, JSONObject event) {
             return -1;
         }
 
-        if (eventType.equals(Constants.IDENTIFY_EVENT)) {
+        if (eventType.equals(Constants.IDENTIFY_EVENT) || eventType.equals(Constants.GROUP_IDENTIFY_EVENT)) {
```
paladin8: I assume we diligently tracked down everywhere `IDENTIFY_EVENT` was used.

src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -1460,7 +1472,29 @@ public void setGroup(String groupType, Object groupName) {
         }
         Identify identify = new Identify().setUserProperty(groupType, groupName);
         logEventAsync(Constants.IDENTIFY_EVENT, null, null, identify.userPropertiesOperations,
-                group, getCurrentTimeMillis(), false);
+                group, null, getCurrentTimeMillis(), false);
+    }
+
+    public void groupIdentify(String groupType, Object groupName, Identify groupIdentify) {
```
paladin8: We call this `groupName` field `groupValue` in the docs: https://amplitude.zendesk.com/hc/en-us/articles/115001765532#group-identify-api. Should probably make them consistent.

Also, why is it `Object` instead of `String`?

djih: I think I made the decision back with the `setGroup` method, since it's sent in a `JSONObject` to allow setting strings or numbers or booleans.

src/com/amplitude/api/AmplitudeClient.java
djih: yep

PR #183, opened 2018-11-29
Handle corrupted db file

djih:
The SDK already deletes the database file on exception and runs the callbacks. Corrupted database files are already handled via the catch exception -> delete / recreate database flow (as verified with unit test). 

Potential edge case that isn't handled: if the database file is deleted by some other process or the OS(?) when the SDK is not interacting with the database, no exception is thrown, and `getWriteableDatabase` actually creates a new file on the next database operation.

We can handle this by also calling the `onDatabaseReset` callback inside `onCreate`, although we have to restructure it because of nested `getWriteableDatabase` calls. Changing the callback to accept a `db` object. This is a little gross though because we're basically exposing the DatabaseHelper internals (like the table constants and another method) to the AmplitudeClient.

Thoughts @curtisliu ? Unclear if this will solve the db file reset issue


src/com/amplitude/api/DatabaseHelper.java
```diff
@@ -96,6 +96,12 @@ public void onCreate(SQLiteDatabase db) {
         // lifetime of the table, even if rows get removed
         db.execSQL(CREATE_EVENTS_TABLE);
         db.execSQL(CREATE_IDENTIFYS_TABLE);
+
+        if (databaseResetListener != null && callResetListenerOnDatabaseReset) {
+            callResetListenerOnDatabaseReset = false;  // guards against stack overflow
+            databaseResetListener.onDatabaseReset(db);
```
curtisliu: if onDatabaseReset fails, callResetListenerOnDatabaseReset will never get reset

djih: kk putting it in a try catch and resetting flag in finally

PR #185, opened 2018-12-14
allow custom server url

djih:


src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -194,6 +194,7 @@ public AmplitudeClient() {
      */
     public AmplitudeClient(String instance) {
         this.instanceName = Utils.normalizeInstanceName(instance);
+        this.url = Constants.EVENT_LOG_URL;
```
paladin8: I kinda like leaving this set when the variable is declared, but either way is fine.

djih: oh that actually makes sense and looks cleaner

Issue #178, opened 2018-11-02
How to track session if a custom ActivityLifecycleCallbacks implementation is already registered


rishabh876:
We already have a custom `ActivityLifecycleCallbacks` implementation registered in our application. In the documentation, I am unable to find a way to accommodate Amplitude's Session Tracking in our app. Usually there are public functions that can be called in such cases like `amplitudeClient.onPause()` and  `amplitudeClient.onResume()` from custom `ActivityLifecycleCallbacks` implementation. But I couldn't find any such workaround for this as well. 
How can I track session in this case?

dantetam:
For you, and anyone else who is interested, we have a guide for this: https://developers.amplitude.com/docs/android#user-sessions

PR #187, opened 2019-01-14
Allow injection of custom Network Client

danmalafaia:
Hello! In the business I work we are required to only make connections using [mTLS](https://en.wikipedia.org/wiki/Mutual_authentication), and, for that reason, we cannot use the SDK as is.

We have an internal proxy server that uses mTLS and talks to the “outside world”, but, in order to use that, we need to inject our custom network client that enforces mTLS and uses our custom request encoding.

What do you think about the current implementation? I made it in a way that it won’t break the existing API. The iOS version of this PR is [iOS #178](https://github.com/amplitude/Amplitude-iOS/pull/178).

### DIFF
--- a/src/com/amplitude/api/DefaultNetworkClient.java
+++ b/src/com/amplitude/api/DefaultNetworkClient.java
@@ -1,0 +1,33 @@
+package com.amplitude.api;
+
+import okhttp3.FormBody;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.Response;
+
+import java.io.IOException;
+
+public class DefaultNetworkClient implements NetworkClient {
+
+    @Override
+    public Response uploadEvents(EventUploadRequest eventUploadRequest, OkHttpClient client) throws IOException {
+
+        String apiVersionString = "" + eventUploadRequest.getApiVersion();
+        String timestampString = "" + eventUploadRequest.getUploadTime();
+
+        FormBody body = new FormBody.Builder()
+                .add("v", apiVersionString)
+                .add("client", eventUploadRequest.getApiKey())
+                .add("e", eventUploadRequest.getEvents())
+                .add("upload_time", timestampString)
+                .add("checksum", eventUploadRequest.getChecksum())
+                .build();
+
+        Request request = new Request.Builder()
+                .url(eventUploadRequest.getUrl())
+                .post(body)
+                .build();
+
+        return client.newCall(request).execute();
+    }
+}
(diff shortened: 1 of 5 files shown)


src/com/amplitude/api/DefaultNetworkClient.java
```diff
@@ -0,0 +1,33 @@
...
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.Response;
+
+import java.io.IOException;
+
+public class DefaultNetworkClient implements NetworkClient {
+
+    @Override
+    public Response uploadEvents(EventUploadRequest eventUploadRequest, OkHttpClient client) throws IOException, IllegalArgumentException {
```
djih: @danmalafaia I notice in your API choice that you are still returning an Okhttp3 response object. In your custom client, are you also going to be using the Okhttp3 library? 

I think it might be cool to let the network client to decide what networking protocol/library to use, although I don't feel strongly either way as long as you're okay using the Okhttp3 api as well. In theory the parent just needs the string response body and int response code from whatever response to do the next part.

src/com/amplitude/api/DefaultNetworkClient.java
```diff
@@ -0,0 +1,33 @@
...
+
+        String apiVersionString = "" + eventUploadRequest.getApiVersion();
+        String timestampString = "" + eventUploadRequest.getUploadTime();
+
+        FormBody body = new FormBody.Builder()
+                .add("v", apiVersionString)
```
djih: I left the same feedback in my review of the iOS version of this PR, but basically it feels weird of this logic to live inside the network client itself since this is just Amplitude's event server API. Regardless of the network client / protocol, the data format should be the same right? It almost feels like the FormBody should be created by the parent and passed into the client already formed, and the client just handles sending it to the proper location / method.

You can ignore all of this feedback if your custom network client has a completely different data format for your own endpoint

src/com/amplitude/api/DefaultNetworkClient.java
danmalafaia: We can create a custom response object not related to Okhttp3 library with only the response code and response body.

However we can't remove the reference from Okhttp3 from the parameters without a big refactor. As the library is already using it by default and in the ssl pinned version it's simpler to keep it.

haoliu-amp:
@danmalafaia, hello, sorry for this item pending for so long, is this still a valid requirement from your side?

PR #188, opened 2019-01-16
Oncorruption logging

djih:
1. Added a custom onCorruption handler. Most of the code is from Android's `DefaultDatabaseErrorHandler`, I just added a call to Diagnostic logger to log an error with the current stack trace
2. I tried to do something fancy where I add a callback function to the corruption handler so it can also try to re-insert the metadata in the onCorruption method, but AFAIK you can't open a new database instance from inside the `DatabaseErrorHandler` class
3. Added some more metrics to the diagnostics logger (free / total / available heap, and free disk space)

Unfortunately was not able to trigger the onCorruption handler in any of the unit tests. For some reason the corrupted files trigger a `SqliteException: Cannot prepare statement, base error code: 26`, instead of `SQLiteDatabaseCorruptException` which triggers the onCorruption path

### DIFF
--- a/src/com/amplitude/api/DatabaseHelper.java
+++ b/src/com/amplitude/api/DatabaseHelper.java
@@ -9,6 +9,7 @@ import android.database.sqlite.SQLiteDoneException;
 import android.database.sqlite.SQLiteException;
 import android.database.sqlite.SQLiteOpenHelper;
 import android.database.sqlite.SQLiteStatement;
+import android.util.Pair;
 
 import org.json.JSONException;
 import org.json.JSONObject;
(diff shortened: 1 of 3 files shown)


src/com/amplitude/api/DatabaseHelper.java
```diff
@@ -21,6 +22,72 @@
 
 class DatabaseHelper extends SQLiteOpenHelper {
 
+    private static class DatabaseCorruptionHandler implements DatabaseErrorHandler {
```
curtisliu: can we just extend DefaultDatabaseErrorHandler and call super in onCorruption?

djih: DefaultDatabaseErrorHandler is a final class, cannot extend :(

djih: Found a way to delegate to an instance of it 👍

djih:
Could delegate to instance of default error handler. PTAL @curtisliu

PR #193, opened 2019-03-20
use shared prefs as backup storage for device id

djih:


src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -359,13 +359,6 @@ public AmplitudeClient setDiagnosticEventMaxCount(int eventMaxCount) {
         return this;
     }
 
-    // this method should only be called from the background log thread
-    private void initializeDeviceInfo() {
```
djih: unused

test/com/amplitude/api/AmplitudeClientTest.java
```diff
@@ -251,56 +262,6 @@ public void testIdentifyMultipleOperations() throws JSONException {
         assertTrue(Utils.compareJSONObjects(userProperties, expected));
     }
 
-    @Test
-    public void testReloadDeviceIdFromDatabase() {
```
djih: Moved these tests into `InitializeTest`

src/com/amplitude/api/Utils.java
```diff
@@ -103,4 +106,20 @@ static String normalizeInstanceName(String instance) {
         }
         return instance.toLowerCase();
     }
+
+    static SharedPreferences getAmplitudeSharedPreferences(Context context, String instanceName) {
+        assert !isEmptyString(instanceName);
+        String prefName = Constants.PACKAGE_NAME + "." + instanceName + "." + context.getPackageName();
+        return context.getSharedPreferences(prefName, Context.MODE_PRIVATE | Context.MODE_MULTI_PROCESS);
```
djih: I think the `Context.MODE_MULTI_PROCESS` syncs the shared preferences across multiple processes, but probably not completely necessary

djih: This is also a single instance shared across all clients so we don't need to cache a pointer to it anywhere (or write our own wrapper)

src/com/amplitude/api/AmplitudeClient.java
```diff
@@ -2087,28 +2080,45 @@ private String initializeDeviceId() {
...
+                Utils.writeStringToSharedPreferences(context, instanceName, DEVICE_ID_KEY, deviceId);
+            }
+
             return deviceId;
         }
 
+        // backup #1: check if device id is stored in shared preferences
+        if (!(Utils.isEmptyString(sharedPrefDeviceId) || invalidIds.contains(sharedPrefDeviceId))) {
+            dbHelper.insertOrReplaceKeyValue(DEVICE_ID_KEY, sharedPrefDeviceId);
+            return sharedPrefDeviceId;
```
paladin8: I think these two cases are correct, but I kind of like the idea of all paths of this method ending up calling `saveDeviceId` just so it's easy to verify that we never put ourselves in an inconsistent state.

src/com/amplitude/api/Utils.java
```diff
@@ -103,4 +106,20 @@ static String normalizeInstanceName(String instance) {
         }
         return instance.toLowerCase();
     }
+
+    static SharedPreferences getAmplitudeSharedPreferences(Context context, String instanceName) {
+        assert !isEmptyString(instanceName);
+        String prefName = Constants.PACKAGE_NAME + "." + instanceName + "." + context.getPackageName();
```
paladin8: I assume this is similar to how we scope the DB data?

src/com/amplitude/api/AmplitudeClient.java
djih: Good suggestion

src/com/amplitude/api/Utils.java
djih: Ya this is how we name our db file: https://github.com/amplitude/Amplitude-Android/blob/master/src/com/amplitude/api/DatabaseHelper.java#L74

This is how we named our legacy shared prefs file: https://github.com/amplitude/Amplitude-Android/blob/master/src/com/amplitude/api/AmplitudeClient.java#L2201-L2219

PR #203, opened 2019-10-01
Privacy guard setting

djih:
Initial pass at implementing a privacy guard setting, basically a quick toggle that turns off the automatic tracking of certain user properties by our SDK (like IP address, city, etc). *TODO:* infosec to finalize the list of properties to include in this setting

This PR is a little longer and more complicated than planned since the tracking options are passed as a JSON object on the event payload to the backend. Privacy guard settings can be implemented as a tracking option with those select fields disabled.

This gets complicated when the customer adds their own tracking options for other fields when enabling privacy guard. We need to merge both tracking options for the final set, although merging is a destructive operation, and therefore we need to maintain the original input tracking option.

I've implemented `mergeIn`, and `copyOf` helper methods to facilitate this. We essentially store the original input tracking option as `inputTrackingOptions`, and then the merge is stored in `appliedTrackingOptions`, which is regenerated whenever the customer sets a new tracking option or enable/disables privacy guard.


src/com/amplitude/api/TrackingOptions.java
```diff
@@ -17,6 +17,14 @@
         Constants.AMP_TRACKING_OPTION_REGION,
     };
 
+    private static String[] PRIVACY_GUARD_PROPERTIES = {
+        Constants.AMP_TRACKING_OPTION_ADID,
```
djih: infosec to finalize this list

PR #206, opened 2019-11-14
Add build.gradle in order to support building sdk from gradle

haoliu-amp:
Lots of source code files are moved but unchanged to follow standards.

Tests:
- Ran gradle build and it worked
- Verified that Ant is not broken by this change


build.gradle
```diff
@@ -0,0 +1,51 @@
...
+    implementation 'com.squareup.okhttp3:okhttp:3.10.0'
+    testImplementation 'com.squareup.okhttp3:mockwebserver:3.10.0'
+    testImplementation 'org.robolectric:robolectric:3.0-rc2'
+    testImplementation 'org.robolectric:shadows-maps:3.0-rc2'
+    testImplementation 'org.powermock:powermock-module-junit4:1.6.2'
+    testImplementation 'org.powermock:powermock-module-junit4-rule:1.6.2'
+    testImplementation 'org.powermock:powermock-api-mockito:1.6.2'
+    testImplementation 'org.powermock:powermock-classloading-xstream:1.6.2'
+    testImplementation 'com.google.android:support-v4:r6'
+}
```
djih: new line?

src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,5 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android" package="com.amplitude">
```
djih: will this affect the distribution of our SDK? when we release to Maven, and when customers pull the SDK, will this affect their experience in any way?

src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,5 @@
...
+
+    <uses-permission android:name="android.permission.INTERNET" />
+
+</manifest>
```
djih: new line?

src/main/AndroidManifest.xml
haoliu-amp: No, it won't. It is specific to us.

src/main/AndroidManifest.xml
haoliu-amp: Usually there's no need to add a new line at the end of the file.

build.gradle
haoliu-amp: Just notice it will give you warning in github if there's no space at the end. Will add it in my next PR.

PR #234, opened 2020-02-25
Able to track `api_level` as a device property

haoliu-amp:
Tested: events went through, not affected.


src/main/java/com/amplitude/api/AmplitudeClient.java
```diff
@@ -991,6 +991,9 @@ protected long logEvent(String eventType, JSONObject eventProperties, JSONObject
             if (appliedTrackingOptions.shouldTrackOsVersion()) {
                 event.put("os_version", replaceWithJSONNull(deviceInfo.getOsVersion()));
             }
+            if (appliedTrackingOptions.shouldTrackApiLevel()) {
+                event.put("api_level", replaceWithJSONNull(Build.VERSION.SDK_INT));
```
djih: this is the android version?

djih: do we need try catch? will `Build.VERSION.SDK_INT` ever cause any exceptions?

PR #235, opened 2020-03-18
Add api to customize library

haoliu-amp:
- Added unit tests.
- Used demo app testing it's working properly.

### DIFF
--- a/src/main/java/com/amplitude/api/AmplitudeClient.java
+++ b/src/main/java/com/amplitude/api/AmplitudeClient.java
@@ -158,6 +158,8 @@ public class AmplitudeClient {
     private boolean trackingSessionEvents = false;
     private boolean inForeground = false;
     private boolean flushEventsOnClose = true;
+    private String libraryName = Constants.LIBRARY;
+    private String libraryVersion = Constants.VERSION;
 
     private AtomicBoolean updateScheduled = new AtomicBoolean(false);
     /**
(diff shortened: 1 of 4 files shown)


src/main/java/com/amplitude/api/AmplitudeClient.java
```diff
@@ -1017,8 +1041,8 @@ protected long logEvent(String eventType, JSONObject eventProperties, JSONObject
             }
 
             JSONObject library = new JSONObject();
-            library.put("name", Constants.LIBRARY);
-            library.put("version", Constants.VERSION);
+            library.put("name", this.libraryName == null ? Constants.LIBRARY_UNKNOWN : this.libraryName);
```
djih: nonblocking suggestion: in the `setLibraryName`, if the input value is `null`, then you can do `this.libraryName = LIBRARY_UNKNOWN`, and then you don't have to do this comparison every time

src/main/java/com/amplitude/api/AmplitudeClient.java
```diff
@@ -1017,8 +1041,8 @@ protected long logEvent(String eventType, JSONObject eventProperties, JSONObject
...
+            library.put("version", this.libraryVersion == null ? Constants.VERSION_UNKNOWN : this.libraryVersion);
```
djih: same here

src/main/java/com/amplitude/unity/plugins/AmplitudePlugin.java
```diff
@@ -94,6 +94,22 @@ public static void enableForegroundTracking(String instanceName, Application app
         Amplitude.getInstance(instanceName).enableForegroundTracking(app);
     }
 
+    public static void enableCoppaControl(String instanceName) {
```
djih: 👍

src/main/java/com/amplitude/api/AmplitudeClient.java
haoliu-amp: My thought is to leave it as is. Keep it original. 

Both are fine.

Issue #240, opened 2020-06-10
Amplitude sending user's location to Google every app launch


liyoung47:
_Initially reported on May 8, 2020_

Every app launch, Amplitude gets the users current location and sends it to Google's reverse geocoding api, and there's no way to prevent this.

`AmplitudeClient.initialize(...)` calls `runOnLogThread()`
the code run on the log thread includes:
```
deviceInfo = new DeviceInfo(context);
deviceInfo.prefetch();
```

`prefetch()` calls  `getCachedInfo();` which calls `new CachedInfo();` The `CachedInfo` constructor calls `getCountry();` which calls `getCountryFromLocation();`

`getCountryFromLocation()` will abort if `DeviceInfo`'s `locationListening` property is false, but there is no way to get it to be false at this point (see below).
If it doesn't abort, `getCountryFromLocation()` then gets the most recent location and sends it to Google.

##### Why disableLocationListening doesn't work:
`AmplitudeClient.disableLocationListening()` sets DeviceInfo's `locationListening` property to false, but it's impossible to get this to happen in time to prevent Amplitude from getting current location and sending to Google.
- If I put the call to `disableLocationListening` before the call to `initialize` then it crashes because AmplitudeClient's `deviceInfo` property is null.
- If I put the call to `disableLocationListening` after the call to `initialize` then that's too late, it already made the call to Google. 

This is the same bug that is causing issue #189 but it was not introduced in PR #117. Even before  `runOnLogThread` was added, `disableLocationListening` wouldn't work unless AmplitudeClient's `deviceInfo` was non-null, and `prefetch()` is called as soon as `deviceInfo` is initialized so there's no way to turn off location listening in time.

BTW I am already calling `AmplitudeClient.setTrackingOptions` and disabling city, country, LatLng, etc before the call to `initialize` and this has no effect.

haoliu-amp:
@liyoung47 Hey, I just released 2.26.1. Should be available from maven repo very soon. 

Please call `disableLocationListening` before `initialize`, and it won't crash now.

Also, if the city and country were not disabled with tracking options, then it will still appear since the previous info will be fetched to use later. However, new users with city, country disabled tracking options, should be able to erase those fields.

PR #241, opened 2020-06-15
Fix - unable to disable location listening at the beginning

haoliu-amp:
Once location is cached, you won't be able to disable location listening at all. However, disabling location listening will result crash if it's called before initialization.
#240

### DIFF
--- a/src/main/java/com/amplitude/api/AmplitudeClient.java
+++ b/src/main/java/com/amplitude/api/AmplitudeClient.java
@@ -270,8 +270,7 @@ public class AmplitudeClient {
                         AmplitudeClient.upgradeSharedPrefsToDB(context);
                     }
                     httpClient = new OkHttpClient();
-                    deviceInfo = new DeviceInfo(context);
-                    deviceInfo.setLocationListening(this.locationListening);
+                    deviceInfo = new DeviceInfo(context, this.locationListening);
                     deviceId = initializeDeviceId();
                     deviceInfo.prefetch();
 
@@ -400,6 +399,9 @@ public class AmplitudeClient {
      */
     public AmplitudeClient enableLocationListening() {
         this.locationListening = true;
+        if (this.deviceInfo != null) {
+            this.deviceInfo.setLocationListening(true);
+        }
         return this;
     }
 
@@ -413,6 +415,9 @@ public class AmplitudeClient {
      */
     public AmplitudeClient disableLocationListening() {
         this.locationListening = false;
+        if (this.deviceInfo != null) {
+            this.deviceInfo.setLocationListening(false);
+        }
         return this;
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/amplitude/api/AmplitudeClient.java
```diff
@@ -270,6 +271,7 @@ public synchronized AmplitudeClient initialize(final Context context, final Stri
                     }
                     httpClient = new OkHttpClient();
                     deviceInfo = new DeviceInfo(context);
+                    deviceInfo.setLocationListening(this.locationListening);
```
djih, line 274: suggestion: you could add `locationListening` as an argument to `DeviceInfo` constructor since you already have it defined here

src/main/java/com/amplitude/api/AmplitudeClient.java
```diff
@@ -392,39 +394,25 @@ public AmplitudeClient useAdvertisingIdForDeviceId() {
...
-            @Override
-            public void run() {
-                if (deviceInfo == null) {
-		    throw new IllegalStateException(
-		            "Must initialize before acting on location listening.");
-                }
-                deviceInfo.setLocationListening(true);
-            }
-        });
+        this.locationListening = true;
```
djih, line 402: don't you need to update locationListening flag in deviceInfo

src/main/java/com/amplitude/api/AmplitudeClient.java
```diff
@@ -392,39 +394,25 @@ public AmplitudeClient useAdvertisingIdForDeviceId() {
...
-            @Override
-            public void run() {
-                if (deviceInfo == null) {
-		    throw new IllegalStateException(
-		            "Must initialize before acting on location listening.");
-                }
-                deviceInfo.setLocationListening(false);
-            }
-        });
+        this.locationListening = false;
```
djih, line 415: don't you need to update locationListening flag in deviceInfo

src/main/java/com/amplitude/api/AmplitudeClient.java
haoliu-amp: Since it only fetches once and then caches the device info, even you have flag set, it won't read it again.

Also this function needs to call before initialization before DeviceInfo fetches the location info.

djih: In logEvent, it always asks DeviceInfo for latest location: https://github.com/amplitude/Amplitude-Android/blob/master/src/main/java/com/amplitude/api/AmplitudeClient.java#L1054.

You might want to check the locationListening flag in AmplitudeClient first before fetching latest location.

Also if the customer disables location listening, but wants to re-enable it, we should probably support that by also updating flag inside Device Info

haoliu-amp: Good catch! Thank you

haoliu-amp: Updated PR.

djih: @haoliu-amp I forgot to mention that this behavior is only in Android SDK. The iOS SDK fetches the location only once, and caches it, and sends it on all events, whereas Android SDK fetches the most recent location every event

haoliu-amp: You mean cached location info inside DeviceInfo?

djih: 👍

PR #275, opened 2021-04-06
feat(Identify): and preInsert, postInser and remove method

yuhao900914:


haoliu-amp (approved):

dantetam (commented):
Little late to the party. Looks good to me. I did a spot check to ensure all the functions have the right signatures and the new test made sense.

PR #276, opened 2021-04-08
Remove okhttp

dantetam:
Do not merge

@haoliu-amp 
I'm sending HTTP requests. I am getting the error response `missing_event` with no extra information. I feel there's just something wrong with the events payload, or there's a different format regarding escaped forward slashes `/`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -64,6 +64,8 @@ android {
 }
 
 dependencies {
+    testImplementation 'com.squareup.okhttp3:okhttp:4.2.2'
+    testImplementation 'com.squareup.okhttp3:mockwebserver:4.2.2'
     testImplementation 'org.robolectric:robolectric:4.3.1'
     testImplementation 'org.robolectric:shadows-maps:3.4-rc2'
     testImplementation 'org.powermock:powermock-module-junit4:1.6.6'
--- a/build.gradle
+++ b/build.gradle
@@ -64,8 +64,6 @@ android {
 }
 
 dependencies {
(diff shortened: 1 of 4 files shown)


src/main/java/com/amplitude/api/AmplitudeClient.java
```diff
@@ -2047,36 +2018,57 @@ protected void makeEventUploadPostRequest(Call.Factory client, String events, fi
...
-            .add("v", apiVersionString)
-            .add("client", apiKey)
-            .add("e", events)
-            .add("upload_time", timestampString)
-            .add("checksum", checksumString)
-            .build();
+        HttpsURLConnection connection = null;
 
-        Request request;
         try {
```
haoliu-amp, line 2023: Better to extract the logic into a separate class.
https://github.com/mixpanel/mixpanel-android/blob/2766eab480eef892a044f7fb2673453919661c11/src/main/java/com/mixpanel/android/util/HttpService.java

dantetam:
Tested on normal Amplitude client and PinnedAmplitudeClient. @qingzhuozhen Do you think the request logic should be in a separate class like @haoliu-amp said?

build.gradle
```diff
@@ -64,7 +64,7 @@ android {
 }
 
 dependencies {
-    implementation 'com.squareup.okhttp3:okhttp:4.2.2'
+    testImplementation 'com.squareup.okhttp3:okhttp:4.2.2'
```
qingzhuozhen, line 67: why we still need this for test?

PR #277, opened 2021-04-20
Quick fix to remove sharedPreferences call/legacy migration

dantetam:
Will handle fixing the tests later. There are too many tests to fix right now and I want to unblock the customer.

### DIFF
--- a/src/main/java/com/amplitude/api/AmplitudeClient.java
+++ b/src/main/java/com/amplitude/api/AmplitudeClient.java
@@ -334,11 +334,6 @@ public class AmplitudeClient {
             if (!initialized) {
                 // this try block is idempotent, so it's safe to retry initialize if failed
                 try {
-                    if (instanceName.equals(Constants.DEFAULT_INSTANCE)) {
-                        AmplitudeClient.upgradePrefs(context);
-                        AmplitudeClient.upgradeSharedPrefsToDB(context);
-                    }
-
                     if (callFactory == null) {
                         // defer OkHttp client to first call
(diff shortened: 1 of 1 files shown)


haoliu-amp (commented):
LGTM. @djih how do you feel?

src/main/java/com/amplitude/api/AmplitudeClient.java
```diff
@@ -334,11 +334,6 @@ public synchronized AmplitudeClient initializeInternal(
             if (!initialized) {
                 // this try block is idempotent, so it's safe to retry initialize if failed
                 try {
-                    if (instanceName.equals(Constants.DEFAULT_INSTANCE)) {
-                        AmplitudeClient.upgradePrefs(context);
```
djih, line 338: do you want to clean up these methods too? `upgradePrefs` and `upgradeSharedPrefsToDB`

djih (approved):

djih:
@haoliu-amp  I think it's fine as long as we're okay breaking backwards compatibility from super legacy versions of Android SDK that are still using sharedPrefs (probably no one is using these versions anymore)

src/main/java/com/amplitude/api/AmplitudeClient.java
dantetam: Going to clean up everything, including the tests @djih

dantetam: In a different branch and PR

PR #290, opened 2021-08-12
Appsetid support

qingzhuozhen:
Add app set id support

### Changes
add app set id get function (Use reflection, tested on emulator with no package installed and package installed, can get the app set id in later case)
support useAppSetIdForDeviceId (Use synchronous to align on the pattern as of today.)
add app set id in api properties
add app set id in tracking options
expose useAppSetIdForDeviceId and disableAppSetId in plugin

[Android doc](https://developer.android.com/training/articles/app-set-id#java)

### DIFF
--- a/src/main/java/com/amplitude/api/AmplitudeClient.java
+++ b/src/main/java/com/amplitude/api/AmplitudeClient.java
@@ -122,6 +122,7 @@ public class AmplitudeClient {
     protected String deviceId;
     private boolean newDeviceIdPerInstall = false;
     private boolean useAdvertisingIdForDeviceId = false;
+    private boolean useAppSetIdForDeviceId = false;
     protected boolean initialized = false;
     private boolean optOut = false;
     private boolean offline = false;
@@ -466,6 +467,18 @@ public class AmplitudeClient {
      */
     public AmplitudeClient useAdvertisingIdForDeviceId() {
         this.useAdvertisingIdForDeviceId = true;
+        this.useAppSetIdForDeviceId = false;
+        return this;
+    }
+
+    /**
+     * Use Android app set id as the user's device ID.
+     *
+     * @return the AmplitudeClient
+     */
+    public AmplitudeClient useAppSetIdForDeviceId() {
+        this.useAppSetIdForDeviceId = true;
+        this.useAdvertisingIdForDeviceId = false;
         return this;
     }
 
@@ -1161,6 +1174,9 @@ public class AmplitudeClient {
             if (appliedTrackingOptions.shouldTrackAdid() && deviceInfo.getAdvertisingId() != null) {
                 apiProperties.put("androidADID", deviceInfo.getAdvertisingId());
             }
+            if (appliedTrackingOptions.shouldTrackAppSetId() && deviceInfo.getAppSetId() != null) {
+                apiProperties.put("androidAppSetId", deviceInfo.getAppSetId());
+            }
             apiProperties.put("limit_ad_tracking", deviceInfo.isLimitAdTrackingEnabled());
             apiProperties.put("gps_enabled", deviceInfo.isGooglePlayServicesEnabled());
 
@@ -1172,7 +1188,6 @@ public class AmplitudeClient {
             event.put("groups", (groups == null) ? new JSONObject() : truncate(groups));
             event.put("group_properties", (groupProperties == null) ? new JSONObject()
                 : truncate(groupProperties));
-
             result = saveEvent(eventType, event);
         } catch (JSONException e) {
             logger.e(TAG, String.format(
@@ -2181,7 +2196,7 @@ public class AmplitudeClient {
 
         // see if device id already stored in db
         String deviceId = dbHelper.getValue(DEVICE_ID_KEY);
-        if (!(Utils.isEmptyString(deviceId) || invalidIds.contains(deviceId))) {
+        if (!(Utils.isEmptyString(deviceId) || invalidIds.contains(deviceId) || deviceId.endsWith("S"))) {
             return deviceId;
         }
 
@@ -2193,6 +2208,15 @@ public class AmplitudeClient {
             if (!(Utils.isEmptyString(advertisingId) || invalidIds.contains(advertisingId))) {
                 saveDeviceId(advertisingId);
                 return advertisingId;
+            }
+        }
+
+        if (useAppSetIdForDeviceId ) {
+            String appSetId = deviceInfo.getAppSetId();
+            if (!(Utils.isEmptyString(appSetId) || invalidIds.contains(appSetId))) {
+                String appSetDeviceId = appSetId + "S";
+                saveDeviceId(appSetDeviceId);
+                return appSetDeviceId;
             }
         }
 
--- a/src/main/java/com/amplitude/api/AmplitudeClient.java
+++ b/src/main/java/com/amplitude/api/AmplitudeClient.java
@@ -121,8 +121,7 @@ public class AmplitudeClient {
      */
     protected String deviceId;
     private boolean newDeviceIdPerInstall = false;
-    private boolean useAdvertisingIdForDeviceId = false;
-    private boolean useAppSetIdForDeviceId = false;
+    private DeviceIdType deviceIdType = null;
     protected boolean initialized = false;
     private boolean optOut = false;
     private boolean offline = false;
@@ -466,8 +465,7 @@ public class AmplitudeClient {
      * @return the AmplitudeClient
      */
     public AmplitudeClient useAdvertisingIdForDeviceId() {
-        this.useAdvertisingIdForDeviceId = true;
-        this.useAppSetIdForDeviceId = false;
+        deviceIdType = DeviceIdType.ADID;
         return this;
     }
 
@@ -477,8 +475,7 @@ public class AmplitudeClient {
      * @return the AmplitudeClient
      */
     public AmplitudeClient useAppSetIdForDeviceId() {
-        this.useAppSetIdForDeviceId = true;
-        this.useAdvertisingIdForDeviceId = false;
+        deviceIdType = DeviceIdType.APP_SET_ID;
         return this;
     }
 
@@ -2200,7 +2197,7 @@ public class AmplitudeClient {
             return deviceId;
         }
 
-        if (!newDeviceIdPerInstall && useAdvertisingIdForDeviceId && !deviceInfo.isLimitAdTrackingEnabled()) {
+        if (!newDeviceIdPerInstall && deviceIdType == DeviceIdType.ADID && !deviceInfo.isLimitAdTrackingEnabled()) {
             // Android ID is deprecated by Google.
             // We are required to use Advertising ID, and respect the advertising ID preference
 
@@ -2211,9 +2208,10 @@ public class AmplitudeClient {
             }
         }
 
-        if (useAppSetIdForDeviceId) {
+        if (deviceIdType == DeviceIdType.APP_SET_ID) {
             String appSetId = deviceInfo.getAppSetId();
             if (!(Utils.isEmptyString(appSetId) || invalidIds.contains(appSetId))) {
+                // Suffix with S for app set id so in future we can tell if device id is from app set id
                 String appSetDeviceId = appSetId + "S";
                 saveDeviceId(appSetDeviceId);
                 return appSetDeviceId;
--- a/src/main/java/com/amplitude/api/AmplitudeClient.java
+++ b/src/main/java/com/amplitude/api/AmplitudeClient.java
@@ -1172,7 +1172,7 @@ public class AmplitudeClient {
                 apiProperties.put("androidADID", deviceInfo.getAdvertisingId());
             }
             if (appliedTrackingOptions.shouldTrackAppSetId() && deviceInfo.getAppSetId() != null) {
-                apiProperties.put("androidAppSetId", deviceInfo.getAppSetId());
+                apiProperties.put("android_app_set_id", deviceInfo.getAppSetId());
             }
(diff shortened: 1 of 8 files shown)


src/main/java/com/amplitude/api/AmplitudeClient.java
```diff
@@ -2181,7 +2196,7 @@ private String initializeDeviceId() {
 
         // see if device id already stored in db
         String deviceId = dbHelper.getValue(DEVICE_ID_KEY);
-        if (!(Utils.isEmptyString(deviceId) || invalidIds.contains(deviceId))) {
+        if (!(Utils.isEmptyString(deviceId) || invalidIds.contains(deviceId) || deviceId.endsWith("S"))) {
```
qingzhuozhen, line 2199: This is based on https://developer.android.com/training/articles/app-set-id#dont-rely-on-cached-value , then we just fetch each time.

src/main/java/com/amplitude/api/AmplitudeClient.java
```diff
@@ -466,6 +467,18 @@ public AmplitudeClient enableNewDeviceIdPerInstall(boolean newDeviceIdPerInstall
...
+        return this;
+    }
+
+    /**
+     * Use Android app set id as the user's device ID.
+     *
+     * @return the AmplitudeClient
+     */
+    public AmplitudeClient useAppSetIdForDeviceId() {
+        this.useAppSetIdForDeviceId = true;
```
dantetam, line 480: Do you think this should be an enum Qingzhuo? We might have more options in the future

src/main/java/com/amplitude/api/AmplitudeClient.java
```diff
@@ -1161,6 +1174,9 @@ protected long logEvent(String eventType, JSONObject eventProperties, JSONObject
             if (appliedTrackingOptions.shouldTrackAdid() && deviceInfo.getAdvertisingId() != null) {
                 apiProperties.put("androidADID", deviceInfo.getAdvertisingId());
             }
+            if (appliedTrackingOptions.shouldTrackAppSetId() && deviceInfo.getAppSetId() != null) {
```
dantetam, line 1177: deviceInfo.getAppSetId() can be set as a variable outside and earlier. For code readability (doesn't actually affect the compiler)

src/main/java/com/amplitude/api/AmplitudeClient.java
```diff
@@ -2196,6 +2211,15 @@ private String initializeDeviceId() {
             }
         }
 
+        if (useAppSetIdForDeviceId ) {
+            String appSetId = deviceInfo.getAppSetId();
+            if (!(Utils.isEmptyString(appSetId) || invalidIds.contains(appSetId))) {
+                String appSetDeviceId = appSetId + "S";
```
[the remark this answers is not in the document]
dantetam: Is the 'S' to mark when a device's ID is an app set ID and therefore should be retrieved again every time the SDK starts up? I don't see any documentation for the letter on Google side. Feel free to leave documentation here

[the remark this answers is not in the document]
dantetam: Leave a code comment is what I meant

src/main/java/com/amplitude/api/DeviceInfo.java
```diff
@@ -200,6 +202,30 @@ private String getAdvertisingId() {
...
+                Method await = Tasks.getMethod("await", Class.forName("com.google.android.gms.tasks.Task"));
+                Object appSetInfo = await.invoke(null, taskWithAppSetInfo);
+                Method getId = appSetInfo.getClass().getMethod("getId");
+                appSetId = (String) getId.invoke(appSetInfo);
+            } catch (ClassNotFoundException e) {
+                AmplitudeLog.getLogger().w(TAG, "Google Play Services SDK not found!");
+            } catch (InvocationTargetException e) {
+                AmplitudeLog.getLogger().w(TAG, "Google Play Services not available");
+            } catch (Exception e) {
+                AmplitudeLog.getLogger().e(TAG, "Encountered an error connecting to Google Play Services for app set id", e);
```
dantetam, line 223: The message in the Advertising ID error should be changed to reflect Ad ID and the new situation around that. Ad ID may not be available for a good reason (user opts out), however, our SDK won't know the exact reason

src/main/java/com/amplitude/api/AmplitudeClient.java
qingzhuozhen: @dantetam Could you elaborate a bit more on the enum idea? Do you mean like a separate variable like `ForDeviceId` to hold a enum for the options?

dantetam: Yes, like that. Having multiple booleans like `useAppSetIdForDeviceId` and `useAdvertisingIdForDeviceId` and so on will get confusing in my opinion

src/main/java/com/amplitude/api/AmplitudeClient.java
```diff
@@ -2196,6 +2211,15 @@ private String initializeDeviceId() {
             }
         }
 
+        if (useAppSetIdForDeviceId ) {
```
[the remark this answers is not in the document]
qingzhuozhen: yeah, let me remove it. I noticed we weren't use any formatter like google formatter for Android repo.

src/main/java/com/amplitude/api/AmplitudeClient.java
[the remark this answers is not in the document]
qingzhuozhen: This is the suffix I picked to tell if in the future what we stored is app set id or not. The corresponding part is in https://github.com/amplitude/Amplitude-Android/pull/290#discussion_r687993695 So later when we init we can tell if it is app set id and we want to fetch again.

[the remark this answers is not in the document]
dantetam: I mean, put the comment text into our code so we remember later on. That makes sense Qingzhuo

src/main/java/com/amplitude/api/AmplitudeClient.java
qingzhuozhen: @dantetam Not quite get `deviceInfo.getAppSetId() can be set as a variable outside and earlier.` Do you mean move it as a variable and define it earlier inside this function?

src/main/java/com/amplitude/api/DeviceInfo.java
qingzhuozhen: Oh, I can add the ad id in the ad id function as well, so we tell the difference from the message

src/main/java/com/amplitude/api/AmplitudeClient.java
dantetam: String appSetId = deviceInfo.getAppSetId();

src/main/java/com/amplitude/api/DeviceInfo.java
qingzhuozhen: just update the messages for ad id and app set id for these error messages, it may be confusing to see these messages when check the log.

src/main/java/com/amplitude/api/AmplitudeClient.java
qingzhuozhen: @dantetam I see your point now, one thing is it may make sense if `appliedTrackingOptions.shouldTrackAppSetId()` is true, then the variable can be used, but if that't not then we don't need a variable for it and it won't be used later in somewhere. If the two if clause order is reverse then it makes more sense as we would always check the app set id. For the code block here, change it or not won't cause too much difference, just some thought for it.

src/main/java/com/amplitude/api/AmplitudeClient.java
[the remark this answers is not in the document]
qingzhuozhen: I see, let me add a comment for it

src/main/java/com/amplitude/api/AmplitudeClient.java
qingzhuozhen: I see, let me replace these two with enum

qingzhuozhen: just replaced to use enum @dantetam

src/main/java/com/amplitude/api/AmplitudeClient.java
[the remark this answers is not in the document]
qingzhuozhen: add a comment for it.

src/main/java/com/amplitude/api/AmplitudeClient.java
```diff
@@ -465,7 +465,17 @@ public AmplitudeClient enableNewDeviceIdPerInstall(boolean newDeviceIdPerInstall
...
+        deviceIdType = DeviceIdType.ADVERTISING_ID;
+        return this;
+    }
+
+    /**
+     * Use Android app set id as the user's device ID.
+     *
+     * @return the AmplitudeClient
+     */
+    public AmplitudeClient useAppSetIdForDeviceId() {
```
haoliu-amp, line 477: Does iOS SDK have this API for idfv? 
If it doesn't, does it make sense to have it here?

qingzhuozhen: iOS sdk is a bit different, it doesn't have an extra interface for this. Instead, it is using the tracking options and directly do a fallback after idfa. Do we want to align the preference by just use app set id as fallback and remove this method? iOS sdk does use both useAdvertisingIdForDeviceId and tracking option to track idfa as well while android doesn't use tracking option for device id judgement.

haoliu-amp: I think this is actually better. You can use your own judgement for it.

src/main/java/com/amplitude/api/AmplitudeClient.java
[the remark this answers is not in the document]
haoliu-amp: Could adid end up with S? It won't include any upper case character?

Btw, is the length fixed for app set id?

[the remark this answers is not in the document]
qingzhuozhen: @haoliu-amp adid now doesn't have any suffix though, and adid is UUID though, so a-f is better not to used as suffix (UUID is case inventive so A-F is also excluded). The output would be normally lowercase but just pick one they won't use for range. And app set id return value is also uuid, for example `368f80ff-404c-48d7-880e-74c5006e373f`

src/main/java/com/amplitude/api/AmplitudeClient.java
qingzhuozhen: Sounds good! If we didn't use tracking options to judge device id, then we can follow the same pattern. It might be better to not have the tracking option and device id logic couple together.

src/main/java/com/amplitude/api/AmplitudeClient.java
```diff
@@ -1161,6 +1171,9 @@ protected long logEvent(String eventType, JSONObject eventProperties, JSONObject
             if (appliedTrackingOptions.shouldTrackAdid() && deviceInfo.getAdvertisingId() != null) {
                 apiProperties.put("androidADID", deviceInfo.getAdvertisingId());
             }
+            if (appliedTrackingOptions.shouldTrackAppSetId() && deviceInfo.getAppSetId() != null) {
+                apiProperties.put("androidAppSetId", deviceInfo.getAppSetId());
```
qingzhuozhen, line 1175: `android_app_set_id`

haoliu-amp: ?

src/main/java/com/amplitude/api/AmplitudeClient.java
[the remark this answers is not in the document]
haoliu-amp: @qingzhuozhen Got it.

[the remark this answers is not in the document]
haoliu-amp: One thing to think about is about the attribution. In the backend, the S needs to be stripped out to enable attribution working.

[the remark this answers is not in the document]
qingzhuozhen: @haoliu-amp I believe they will use the `android_app_set_id` field in the event api properties to do the attribution?  That field won't have a suffix with it. But I will let them know the device id with app set id will have a suffix with it.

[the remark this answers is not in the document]
haoliu-amp: I know S is to identify the source, but how can it be useful?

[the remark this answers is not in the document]
qingzhuozhen: @haoliu-amp The only purpose to identify the source is to re-fetch the app set id each time when we initialize, in case the app set id get updated. https://developer.android.com/training/articles/app-set-id#dont-rely-on-cached-value If we believe we don't need to re-fetch app set id for device id, then I can remove the suffix and the check to see if it is app set id. Just adding now is the best chance if we want to enable a re-fetch on app set id.

src/main/java/com/amplitude/api/AmplitudeClient.java
qingzhuozhen: @haoliu-amp this is per the discussion earlier, Kurt mentioned the attributes in backend is mostly with underscore, so just change the label to align it.

src/main/java/com/amplitude/api/AmplitudeClient.java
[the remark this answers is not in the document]
haoliu-amp: Got it. Keep it that way.

haoliu-amp (approved):

PR #291, opened 2021-08-23
App Set ID tests

dantetam:
New tests for app set id enum/flag, sending app set ID as event JSON payload

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -75,6 +75,7 @@ dependencies {
     testImplementation 'com.google.android:support-v4:r6'
     testImplementation 'com.google.android.gms:play-services-ads:18.3.0'
     testImplementation 'com.google.android.gms:play-services-base:17.1.0'
+    testImplementation 'org.json:json:20140107'
 
     testImplementation "junit:junit:4.12"
 
--- a/src/test/java/com/amplitude/api/DeviceInfoTest.java
+++ b/src/test/java/com/amplitude/api/DeviceInfoTest.java
@@ -13,16 +13,22 @@ import androidx.test.ext.junit.runners.AndroidJUnit4;
 import com.google.android.gms.ads.identifier.AdvertisingIdClient;
 import com.google.android.gms.common.ConnectionResult;
 import com.google.android.gms.common.GooglePlayServicesUtil;
+import com.google.android.gms.tasks.Task;
+import com.google.android.gms.tasks.Tasks;
 
+import org.json.JSONObject;
 import org.junit.After;
+import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
+import org.mockito.Mock;
 import org.mockito.Mockito;
 import org.powermock.api.mockito.PowerMockito;
 import org.powermock.core.classloader.annotations.PowerMockIgnore;
 import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
 import org.powermock.modules.junit4.rule.PowerMockRule;
 import org.robolectric.Robolectric;
 import org.robolectric.Shadows;
@@ -32,18 +38,20 @@ import org.robolectric.shadows.ShadowTelephonyManager;
 import org.robolectric.util.ReflectionHelpers;
 
 import java.util.Locale;
+import java.util.UUID;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 
 @RunWith(AndroidJUnit4.class)
-@PowerMockIgnore({ "org.mockito.*", "org.robolectric.*", "android.*", "androidx.*", "javax.net.ssl.*" })
-@PrepareForTest({AdvertisingIdClient.class, GooglePlayServicesUtil.class})
+@PowerMockIgnore({ "org.mockito.*", "org.robolectric.*", "android.*", "androidx.*", "javax.net.ssl.*", "jdk.internal.reflect.*", "javax.management.*" })
+@PrepareForTest({AdvertisingIdClient.class, GooglePlayServicesUtil.class, UUID.class, DeviceInfo.class})
 @Config(manifest = Config.NONE)
 public class DeviceInfoTest extends BaseTest {
     private DeviceInfo deviceInfo;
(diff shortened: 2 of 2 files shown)


build.gradle
```diff
@@ -75,6 +75,7 @@ dependencies {
     testImplementation 'com.google.android:support-v4:r6'
     testImplementation 'com.google.android.gms:play-services-ads:18.3.0'
     testImplementation 'com.google.android.gms:play-services-base:17.1.0'
+    testImplementation 'org.json:json:20140107'
```
dantetam, line 78: https://stackoverflow.com/questions/29402155/android-unit-test-not-mocked

src/test/java/com/amplitude/api/DeviceInfoTest.java
```diff
@@ -304,4 +312,90 @@ public void testDontUseAdvertisingIdAsDeviceId() {
...
+
+        AmplitudeClient client = Amplitude.getInstance("UseAppSetId");
+        client.useAppSetIdForDeviceId();
+        client.initialize(context, "1cc2c1978ebab0f6451112a8f5df4f4e");
+        ShadowLooper looper = Shadows.shadowOf(client.logThread.getLooper());
+        looper.runToEndOfTasks();
+
+        assertNotEquals(advertisingId, client.getDeviceId());
+        assertNotEquals(mockUUIDString, client.getDeviceId());
+        assertTrue(client.getDeviceId().endsWith("S"));
```
haoliu-amp, line 353: Are we expecting that deviceId equals to appSetId? Need to test if it equals to "5a8f0fd1-31a9-4a1f-bfad-cd5439ce533b" + "S"

src/test/java/com/amplitude/api/DeviceInfoTest.java
```diff
@@ -304,4 +312,90 @@ public void testDontUseAdvertisingIdAsDeviceId() {
         assertNotEquals(advertisingId, client.getDeviceId());
         assertTrue(client.getDeviceId().endsWith("R"));
     }
+
+    @Test
+    public void testAppSetId() {
```
haoliu-amp, line 317: Rename it to `testDeviceIdEqualsToAppSetId`. Make it descriptive.

haoliu-amp (changes requested):

src/test/java/com/amplitude/api/DeviceInfoTest.java
```diff
@@ -304,4 +312,90 @@ public void testDontUseAdvertisingIdAsDeviceId() {
...
+
+        Robolectric.getForegroundThreadScheduler().advanceTo(1);
+
+        AmplitudeClient client = Amplitude.getInstance("UseAppSetId");
+        client.useAppSetIdForDeviceId();
+        client.initialize(context, "1cc2c1978ebab0f6451112a8f5df4f4e");
+        ShadowLooper looper = Shadows.shadowOf(client.logThread.getLooper());
+        looper.runToEndOfTasks();
+
+        assertNotEquals(advertisingId, client.getDeviceId());
```
qingzhuozhen, line 351: Why we are asserting like this? Why not just assertEquals mockAppsetId + "S"? Then we don't really need mock other parts for ad id or uuid right?

dantetam: That's a good point. I just wanted to ensure the device id used was not some other ID. It's an unnecessary mock and assert.

src/test/java/com/amplitude/api/DeviceInfoTest.java
```diff
@@ -304,4 +312,90 @@ public void testDontUseAdvertisingIdAsDeviceId() {
...
+        String mockAppSetId = "5a8f0fd1-31a9-4a1f-bfad-cd5439ce533b";
+        PowerMockito.stub(PowerMockito.method(DeviceInfo.class, "getAppSetId"))
+                .toReturn(mockAppSetId);
+
+        ShadowLooper looper = Shadows.shadowOf(amplitude.logThread.getLooper());
+
+        amplitude.useAppSetIdForDeviceId();
+        amplitude.initialize(context, apiKey);
+        looper.runToEndOfTasks();
+        assertEquals(mockAppSetId + "S", amplitude.getDeviceId());
```
qingzhuozhen, line 367: Actually to this line in testDoNotSendAppSetId test case is basically what we need in last test case

src/test/java/com/amplitude/api/DeviceInfoTest.java
```diff
@@ -304,4 +312,90 @@ public void testDontUseAdvertisingIdAsDeviceId() {
...
+
+        JSONObject event = getLastEvent();
+        assertNotNull(event);
+        try {
+            assertEquals("testSendAppSetIdInJson", event.getString("event_type"));
+            JSONObject apiProps = event.getJSONObject("api_properties");
+            String appSetId = apiProps.getString("android_app_set_id");
+            assertEquals(mockAppSetId, appSetId);
+        } catch (Exception e) {
+            e.printStackTrace();
```
qingzhuozhen, line 380: I guess we can do `fail(e.toString());` as we have now for the exception

build.gradle
qingzhuozhen: I believe we do use JSONObject in other test files, why this time there would be errors?

dantetam: Could you try running these tests locally @qingzhuozhen ? This is one of the issues we may solve with multiple devices

qingzhuozhen: yeah, I can run the tests locally. I guess the issue is for in the test case `testDoNotSendAppSetId` we use PowerMock there. I wonder if we should adjust our code a bit so we don't need to use PowerMock there and just use Mockito, then we don't really need to import the lib into our tests.

src/test/java/com/amplitude/api/DeviceInfoTest.java
```diff
@@ -304,4 +312,90 @@ public void testDontUseAdvertisingIdAsDeviceId() {
...
+
+        assertNotEquals(advertisingId, client.getDeviceId());
+        assertNotEquals(mockUUIDString, client.getDeviceId());
+        assertTrue(client.getDeviceId().endsWith("S"));
+    }
+
+    @Test
+    public void testDoNotSendAppSetId() {
+        String mockAppSetId = "5a8f0fd1-31a9-4a1f-bfad-cd5439ce533b";
+        PowerMockito.stub(PowerMockito.method(DeviceInfo.class, "getAppSetId"))
```
qingzhuozhen, line 359: This line seems the place causing issues and we have to import the lib? What if when we initialize we create a new method `collectDeviceInfo` from our client, then we can mock this method on amplitude interface to return a mocked DeviceInfo object, then we can just use Mockito instead of PowerMock in this test case. And if that works, we don't need to import the lib.

src/test/java/com/amplitude/api/DeviceInfoTest.java
qingzhuozhen: If we assert the equals part then it should be enough, and I believe the mocks are not really got hit.

src/test/java/com/amplitude/api/DeviceInfoTest.java
dantetam: That's an interesting idea here. I'll try to implement on my end

dantetam: I considered doing this, but I didn't want to touch any private functions or change the encapsulation. My understanding is that if a method is private, it shouldn't be involved as the main focus of a test. 

I'm totally fine with just importing `json...` as a `testImplementation` dependency. Our code does touch the `android json` library, and from stackoverflow, mockito tests of any kind should not rely on built-in Android libraries. It's their strange rule, you can't involve others' code in your test where you are mocking your own code.

qingzhuozhen: I guess it might be fine to leave it like this, just for best practice we should try to avoid that as possible. Also as mocking happening on DeviceInfo, and there is no JSONObject usage inside that class, it is a bit confusing to adding in JSONObject as part of test dependency. For better practice, we may want to avoid extra dependency and much PowerMock usage.

PR #298, opened 2021-09-07
feat: add optional error callback support

qingzhuozhen:


src/main/java/com/amplitude/api/AmplitudeErrorCallback.java
```diff
@@ -0,0 +1,5 @@
+package com.amplitude.api;
+
+public interface AmplitudeErrorCallback {
```
haoliu-amp, line 3: Maybe rename this to `AmplitudeLogCallback`, so that you can reuse this class later.

dantetam: Oh I was going to ask about this, why are we using a custom Amplitude interface instead of the Android standardized callback class? I can see benefits to either approach

dantetam: It's this class [here](https://developer.android.com/reference/javax/security/auth/callback/Callback) . Of course, not necessary if you want to move forward with this solution

qingzhuozhen: That seems just an empty interface. I believe we don't need to rely on extra dependency for our SDK.

dantetam: Fair enough. I was confusing it with this [class](https://developer.android.com/reference/android/os/Handler.Callback). This is more Android specific related to the Handler paradigm. Let's move forward with this approach

haoliu-amp (approved):

Issue #282, opened 2021-04-30
deviceId is null after call initialize()


krottv:
## Expected Behavior
deviceId shouldn't be null or we need to have a callback to know when amplitude is initialized correctly.

## Current Behavior
Hi. For integration for other analytics tools I need to know amplitude deviceId. And it is null after Amplitude.initialize(...) call. But If I fetch deviceId after like a second after initialize call then it is non null. 

## Possible Solution
Make a delay before fetching deviceId, but the problem is that I don't know in advance how long the initialization will take.

## Steps to Reproduce
 

1. Amplitude.initialize(...) 
2. Amplitude.getInstance().deviceId == null

## Environment
- SDK Version: 2.31.0
- Android API Level: 29
- Device: Samsung Galaxy s9+

krottv:
@dantetam 
As I see from the source code there are two solutions and they seem pretty easy: 1. initialize deviceId instantly in the caller thread. or 2. add a callback when initialization is finished.

dantetam:
Hello @krottv , 

We're going to close this issue because it's stale. If you need a deterministic (i.e. 100% certain solution) for a device ID, feel free to wait a long time or generate a device ID yourself and set it using the method below:

```
/**
     * Sets a custom device id. <b>Note: only do this if you know what you are doing!</b>
     *
     * @param deviceId the device id
     * @return the AmplitudeClient
     */
    public AmplitudeClient setDeviceId(final String deviceId) { ... }
```

dantetam:
Hello @krottv and anyone else listening,

We plan to take this on very soon. Same for iOS. Thank you for the input.

krottv:
@dantetam Nice.
The usecase is following. Inside the app I use Adapty - subscription management service which integrates with Amplitude. And in order to make integration, I need to pass amplitude.deviceId to Adapty. Here's docs for more info - https://doc.adapty.io/docs/amplitude

dantetam:
Sorry for the late delay on this @krottv . Are you by chance using any workarounds right now? Like just waiting for a while? We've created a feature request for this ticket

krottv:
@dantetam Yes. I use delay(500)

PR #301, opened 2021-09-22
Add device id callback

dantetam:


dantetam:
Fixes #282 Sorry @krottv for the delay

dantetam:
Also this is good future proofing because the Experiment team had asked about iOS

src/main/java/com/amplitude/api/AmplitudeClient.java
```diff
@@ -357,6 +358,10 @@ public void onFinished() {
                     deviceId = initializeDeviceId();
                     deviceInfo.prefetch();
 
+                    if (this.deviceIdCallback != null) {
```
qingzhuozhen, line 361: LGTM, one quick questions is do we just care about the deviceId here? I wonder if we should move this before the `prefetch` method? I guess that method would make calls and causing some latency.

dantetam: Oh yeah, good catch.

qingzhuozhen (approved):

PR #308, opened 2021-12-09
feat: add support for middleware

yuhao900914:


src/main/java/com/amplitude/api/AmplitudeClient.java
```diff
@@ -201,6 +201,10 @@
      * The background event uploading worker thread instance.
      */
     WorkerThread httpThread = new WorkerThread("httpThread");
+    /**
+     * The runner for middleware
+     * */
+    MiddllewareRunner middlewareRunner = new MiddllewareRunner();
```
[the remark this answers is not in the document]
justin-fiedler: You will also need to expose a `addEventMiddleware` method on the client to allow users to add middleware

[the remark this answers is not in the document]
yuhao900914: Yes, will come with the next commit.

justin-fiedler:
Can you also please add `extra` to `setUserProperties` we are using it in codegen since it's more concise than constructing an Identfy object from JSON. 
https://github.com/amplitude/javascript/pull/13541/files#diff-fba79090504d6bb2a35fa69573380efd599f48f048d825c60c6ada66375c2305R246

src/main/java/com/amplitude/api/AmplitudeClient.java
```diff
@@ -1640,14 +1682,14 @@ public void identify(Identify identify) {
      * @param identify an {@link Identify} object
      * @param outOfSession whther to log the identify event out of session
      */
-    public void identify(Identify identify, boolean outOfSession) {
+    public void identify(Identify identify, boolean outOfSession, MiddlewareExtra extra) {
```
justin-fiedler, line 1685: Can you also please add extra to `setUserProperties()` we are using it in codegen since it's more concise than constructing an Identfy object from JSON.
https://github.com/amplitude/javascript/pull/13541/files#diff-fba79090504d6bb2a35fa69573380efd599f48f048d825c60c6ada66375c2305R246

justin-fiedler (changes requested):

src/test/java/com/amplitude/api/MiddlewareTest.java
```diff
@@ -0,0 +1,160 @@
...
+import static org.robolectric.Shadows.shadowOf;
+
+@RunWith(AndroidJUnit4.class)
+@Config(manifest= Config.NONE)
+public class MiddlewareTest extends BaseTest {
+
+    @Before
+    public void setUp() throws Exception {
+        super.setUp();
+        amplitude.initialize(context, apiKey);
```
qingzhuozhen, line 34: Ideally unit test here can limit to the class itself. For example, this file just covers Middleware but not the amplitude client. And in amplitude client we cover from amplitude test and how ever looks like. So if anything wrong it is easier to identify the issue. It may look verbose for this case. However I believe that is a better practice in general.

src/test/java/com/amplitude/api/MiddlewareTest.java
```diff
@@ -0,0 +1,160 @@
...
+        Shadows.shadowOf(amplitude.logThread.getLooper()).runOneTask();
+    }
+
+    @After
+    public void tearDown() throws Exception {
+        super.tearDown();
+    }
+
+    private MiddlewareNext next = new MiddlewareNext() {
```
qingzhuozhen, line 43: Is this used somewhere?

src/main/java/com/amplitude/api/MiddlewareRunner.java
```diff
@@ -0,0 +1,33 @@
...
+            @Override
+            public void run(MiddlewarePayload curPayload) {
+                runMiddlewares(middlewares.subList(1, middlewares.size()), curPayload, next);
+            }
+        });
+    }
+
+    public void run(MiddlewarePayload payload, MiddlewareNext next) {
+        runMiddlewares(this.middlewares, payload, next);
+    }
```
justin-fiedler, line 32:
```
public boolean run(MiddlewarePayload payload) {
    AtomicBoolean middlewareComplemeted = false;
    this.run(new MiddlewarePayload(event, extra),  new MiddlewareNext() {
        @Override
        public void run(MiddlewarePayload curPayload) {
            middlewareComplemeted = true;
        }
    });

    return middlewareComplemeted;
```

src/main/java/com/amplitude/api/AmplitudeClient.java
```diff
@@ -1610,7 +1669,7 @@ public void setUserProperties(final JSONObject userProperties) {
                 logger.e(TAG, e.toString());
             }
         }
-        identify(identify);
+        identify(identify,false, extra);
```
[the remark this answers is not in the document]
justin-fiedler: @yuhao900914

justin-fiedler (changes requested):

justin-fiedler (approved):

qingzhuozhen (approved):

PR #319, opened 2022-02-13
fixed detecting device language

AlexKorRnd:
In case of changing the app language programmatically it also needs to call `Locale.setDefault()` because `Locale.getDefault()` is  used in many android SDK places. So it's better to use `Resources.getSystem().getConfiguration().getLocales() ` to detect the device language.

### DIFF
--- a/src/main/java/com/amplitude/api/DeviceInfo.java
+++ b/src/main/java/com/amplitude/api/DeviceInfo.java
@@ -4,11 +4,14 @@ import android.content.ContentResolver;
 import android.content.Context;
 import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager.NameNotFoundException;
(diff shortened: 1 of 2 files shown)


justin-fiedler:
Hi @AlexKorRnd thank you for the PR. Changes look good but when running test locally I get an error in `DeviceInfo.testGetCountry`. Are you able to run the tests locally and address the issue?

image

justin-fiedler (changes requested):
Please fix broken test.

AlexKorRnd:
@justin-fiedler take a look the new commit [645c76d](https://github.com/amplitude/Amplitude-Android/pull/319/commits/645c76da911f7a18ab449136fa595a438f04fc79)

yuhao900914 (approved):

PR #324, opened 2022-02-21
add groupIdentify overload to set multiple group properties

falconandy:


src/main/java/com/amplitude/api/AmplitudeClient.java
```diff
@@ -1795,7 +1806,14 @@ public void groupIdentify(String groupType, Object groupName, Identify groupIden
     }
 
     public void groupIdentify(String groupType, Object groupName, Identify groupIdentify, boolean outOfSession) {
-        groupIdentify(groupType, groupName, groupIdentify, false, null);
+        groupIdentify(groupType, groupName, groupIdentify, outOfSession, null);
+    }
+
+    public void groupIdentify(String groupType, Object groupName, JSONObject groupProperties, boolean outOfSession, MiddlewareExtra extra) {
```
qingzhuozhen, line 1812: nit: Seems we don't have docs for any groupIdentify method. Not an issue for this PR though. Would be ideal if we can have some.

qingzhuozhen (approved):

justin-fiedler (approved):

src/main/java/com/amplitude/api/AmplitudeClient.java
```diff
@@ -1814,7 +1832,7 @@ public void groupIdentify(String groupType, Object groupName, Identify groupIden
 
         logEventAsync(
             Constants.GROUP_IDENTIFY_EVENT, null, null, null, group,
-            groupIdentify.userPropertiesOperations, getCurrentTimeMillis(), outOfSession
+            groupIdentify.userPropertiesOperations, getCurrentTimeMillis(), outOfSession, extra
```
justin-fiedler, line 1835: 🙏

PR #332, opened 2022-04-19
custom advertising id

vishna:
Solves https://github.com/amplitude/Amplitude-Flutter/issues/101

### DIFF
--- a/src/main/java/com/amplitude/api/DeviceInfo.java
+++ b/src/main/java/com/amplitude/api/DeviceInfo.java
@@ -36,6 +36,8 @@ public class DeviceInfo {
 
     private CachedInfo cachedInfo;
 
+    public static String customAdvertisingId;
+
     /**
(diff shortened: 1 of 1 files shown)


yuhao900914:
Hi @vishna,  thanks for contributing our sdks!
Can you also add a test for this change?

src/main/java/com/amplitude/api/DeviceInfo.java
```diff
@@ -194,6 +196,9 @@ private String getLanguage() {
         }
 
         private String getAdvertisingId() {
+            if (DeviceInfo.customAdvertisingId != null) {
+                return DeviceInfo.customAdvertisingId;
```
qingzhuozhen, line 200: Thanks for the contribution! Maybe this could be on client level other than in DeviceInfo? Multiple instances may cause conflict I guess.

PR #337, opened 2022-07-15
feat: enable eu ssl pinning

yuhao900914:
Summary:
Based on #jpmc-certificate-pinning channel we are going to support SSL pinning on EU. 
And we decided to pin all AWS Root CA for EU requests. 
The EU cert Public Key id comes from https://www.amazontrust.com/repository/, which is used by EU endpoint.

Tests
Unit tests
Event observed in EU Amplitude with SSL pinning enabled
Event observed in US Amplitude with SSL pinning enabled
With wrong setup order, initialize with EU API key -> logEvent -> setServerZone, SSL certificated will be matched after set the setServerZone.


src/main/java/com/amplitude/api/PinnedAmplitudeClient.java
```diff
@@ -42,7 +42,7 @@ public class PinnedAmplitudeClient extends AmplitudeClient {
 
     // CN=COMODO RSA Domain Validation Secure Server CA, O=COMODO CA Limited,
     // L=Salford, ST=Greater Manchester, C=GB
-    private static final String CERTIFICATE_1 = ""
```
justin-fiedler, line 45: 🙏

justin-fiedler (approved):
LGTM

src/main/java/com/amplitude/api/PinnedAmplitudeClient.java
```diff
@@ -286,11 +322,24 @@ protected SSLSocketFactory getPinnedCertSslSocketFactory(SSLContextBuilder conte
...
         }
         return sslSocketFactory;
     }
+
+    @Override
+    public AmplitudeClient setServerZone(AmplitudeServerZone serverZone) {
+        super.setServerZone(serverZone);
+        this.initializedSSLSocketFactory = false;
+        this.sslSocketFactory = null;
+        this.initialize(this.context, this.apiKey, this.userId);
```
qingzhuozhen, line 342: For cal initialize, would there be any other thing affected? Like config, user id, device id, etc.? Also if there is no api key, would it be fine?

yuhao900914: It should be fine.
1. in iOS, we saved the userId and deviceId in the DB. If we can get the value from the DB, we will not re-initialize it. Other configurations are not overwritten in the initialize step. 
2. It will be fine. It will be the same error as the first initialization.

src/test/java/com/amplitude/api/PinningTest.java
```diff
@@ -34,14 +34,53 @@ public void tearDown() throws Exception {
...
+        httplooper.runToEndOfTasks();
+
+        assertNull(amplitude.lastError);
+    }
+
+    @Test
+    public void testSslPinningEU() {
+        amplitude = PinnedAmplitudeClient.getInstance();
+        amplitude.setServerZone(AmplitudeServerZone.EU);
+        amplitude.initialize(context, "361e4558bb359e288ef75d1ae31437a0");
```
liuyang1520, line 58: Is this the real integration test Android too?

liuyang1520 (approved):
LGTM!

qingzhuozhen (approved):

PR #342, opened 2022-08-30
feat: add ingestion_metadata field

liuyang1520:
### Summary
- feat: add ingestion_metadata field

Add this option in the configuration, which can be used in the library wrapper/code generation/dynamic loading cases, for setting the `ingestion_metadata` information.

### DIFF
--- a/src/main/java/com/amplitude/api/Constants.java
+++ b/src/main/java/com/amplitude/api/Constants.java
@@ -88,6 +88,6 @@ public class Constants {
     public static final String AMP_PLAN_VERSION = "version";
     public static final String AMP_PLAN_VERSION_ID = "versionId";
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/amplitude/api/Constants.java
```diff
@@ -87,4 +87,7 @@ public class Constants {
     public static final String AMP_PLAN_SOURCE = "source";
     public static final String AMP_PLAN_VERSION = "version";
     public static final String AMP_PLAN_VERSION_ID = "versionId";
+
+    public static final String INGESTION_METADATA_SOURCE_NAME = "source_name";
```
justin-fiedler, line 91: nit: use `AMP_` prefix to match other consts?

justin-fiedler (approved):

Issue #348, opened 2023-01-27
AmplitudeClient.makeEventUploadPostRequest shall check HTTP response 200 rather than matching string response to "success"


jackcsk:
## Summary

In related to the `amplitude-javascript` pull request https://github.com/amplitude/Amplitude-JavaScript/pull/572, `AmplitudeClient.makeEventUploadPostRequest(client, event, maxEventId, maxIdentifyId)` shall match `Response.code` to 200 rather than doing a string comparison of `"success"`

Current behavior blocks the usage of custom amplitude proxy, which returns HTTP 200 response in this format:
```
{"code":200,"status":"SUCCESS","data":true}
```

## Error Message
From logcat:
```
01-27 19:17:55.455 17721 17910 W com.amplitude.api.AmplitudeClient: Upload failed, {"code":200,"status":"SUCCESS","data":true}
01-27 19:17:55.455 17721 17910 W com.amplitude.api.AmplitudeClient: , will attempt to reupload later
```

jackcsk:
@yuhao900914 Please note that these 2 issues are related, but #349 doesn't solve this problem as they are on different platforms.
Please keep this issue open, thanks

PR #349, opened 2023-01-30
fix: check response code when response code is 200

yuhao900914:
When making an event upload request, check the response code === 200 instead of checking the string response, in order to support the usage of a custom amplitude proxy.

https://github.com/amplitude/Amplitude-Android/issues/348
https://github.com/amplitude/Amplitude-JavaScript/pull/572


liuyang1520 (approved):
Change looks good to me, thanks for fixing this! Please fix the gradle build failure if possible.

liuyang1520 (approved):

PR #354, opened 2023-02-17
fix: catch the case in intercept for identity update

qingzhuozhen:
#### Summary

When user id or device id update, transfer the identify events intercepted to prevent information mix.

### DIFF
--- a/src/main/java/com/amplitude/api/IdentifyInterceptor.java
+++ b/src/main/java/com/amplitude/api/IdentifyInterceptor.java
@@ -27,6 +27,10 @@ class IdentifyInterceptor {
 
     private final AmplitudeClient client;
 
+    private String userId;
+    private String deviceId;
+    private final AtomicBoolean identitySet = new AtomicBoolean(false);
+
     public  IdentifyInterceptor (
             DatabaseHelper dbHelper,
             WorkerThread logThread,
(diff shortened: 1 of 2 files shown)


src/main/java/com/amplitude/api/IdentifyInterceptor.java
```diff
@@ -206,4 +214,36 @@ private boolean isActionOnly(JSONObject event, String action) {
...
+                userId = event.getString("user_id");
+                isUpdated = true;
+            }
+            if (isIdUpdated(deviceId, event.getString("device_id"))) {
+                deviceId = event.getString("device_id");
+                isUpdated = true;
+            }
+            return isUpdated;
+        } catch (JSONException e) {
+            return false;
```
izaaz, line 236: Default to true?

src/test/java/com/amplitude/api/AmplitudeClientTest.java
```diff
@@ -2493,4 +2493,63 @@ public void testMultipleIdentifyWithSetActionAndSetGroup() throws JSONException
...
+        assertEquals(event.getLong("sequence_number"), 1);
+        assertEquals(event.getString("user_id"), "null");
+        JSONObject userProperties = event.getJSONObject("user_properties");
+        assertEquals(userProperties.length(), 1);
+        JSONObject expected = new JSONObject();
+        expected.put("key1", "key1-value2");
+        expected.put("key2", "key2-value2");
+        expected.put("key3", "key3-value2");
+        expected.put("key4", "key4-value2");
+        JSONObject expectedGroups = new JSONObject();
```
izaaz, line 2535: Is `expectedGroups` being used anywhere in the tests?

izaaz (approved):

src/test/java/com/amplitude/api/AmplitudeClientTest.java
qingzhuozhen: nvm, that's a copy from other test, cleaning it up

src/main/java/com/amplitude/api/IdentifyInterceptor.java
qingzhuozhen: yeah, I guess true is safer. Let me update it.

PR #357, opened 2023-04-19
fix: update identify interceptor to identify only

qingzhuozhen:
### Summary

- update to intercept identify only


justin-fiedler (approved):
Look good. Thanks @qingzhuozhen

justin-fiedler (changes requested):
Just released current changes only cover standard events. I think we need to do the same for "active" idenitfies on line 73 of IdentifyInterceptor.

justin-fiedler (approved):

PR #362, opened 2023-05-31
fix: move inForeground reads/writes to main thread

falconandy:
inForeground reads/writes moved to main thread.
The session logic has some differences compared to Amplitude-iOS. I would re-write session logic to match Ampltude-Kotlin/Amplitude-Swift implementation.
For some reason I have a few issues with the Amplitude-Android project on my machine with latest Android Studio:  unresolved identifiers in test files, `./gradlew build` fails with `204 tests completed, 35 failed` (but in Android Studio UI all tests are green).

### DIFF
--- a/src/main/java/com/amplitude/api/AmplitudeClient.java
+++ b/src/main/java/com/amplitude/api/AmplitudeClient.java
@@ -1102,7 +1102,7 @@ public class AmplitudeClient {
      */
     public void logEventSync(String eventType, JSONObject eventProperties, JSONObject groups, long timestamp, boolean outOfSession) {
         if (validateLogEvent(eventType)) {
-            logEvent(eventType, eventProperties, null, null, groups, null, timestamp, outOfSession);
+            logEvent(eventType, eventProperties, null, null, groups, null, timestamp, outOfSession, this.inForeground);
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/amplitude/api/AmplitudeClient.java
```diff
@@ -1618,7 +1621,6 @@ public void onFinished() {
                     }, serverZone);
                 }
                 startNewSessionIfNeeded(timestamp);
```
justin-fiedler, line 1623: Thanks @falconandy, I see what you mean about the logic being different. Event before your changes the call to `startNewSessionIfNeeded(timestamp)` should be working correctly even if `onEnterBackground` isn't called preventing long sessions (it assumes you're in the background).

Not sure I understand how the threading changes will make a difference. Can you please explain more. Thanks

falconandy: > Event before your changes the call to startNewSessionIfNeeded(timestamp) should be working correctly even if onEnterBackground isn't called preventing long sessions (it assumes you're in the background).

Yes, you are right.

Before my changes data races were possible (`inForeground` was set in log thread and can be read in main/log thread). Now all writes/reads are moved to main thread.

justin-fiedler (approved):

justin-fiedler (approved):

justin-fiedler (approved):

PR #366, opened 2023-06-09
fix: do not fetch advertising Id if adid tracking is disabled

falconandy:
Additional note: in Android-Kotlin default value for `limitAdTrackingEnabled` is `true`, in Amplitude-Android - `false`
* https://github.com/amplitude/Amplitude-Kotlin/blob/main/common-android/src/main/java/com/amplitude/common/android/AndroidContextProvider.kt#L51
* https://github.com/amplitude/Amplitude-Android/blob/main/src/main/java/com/amplitude/api/DeviceInfo.java#L56

Maybe if adid tracking is disabled `limitAdTrackingEnabled` should be `null`.

### DIFF
--- a/src/main/java/com/amplitude/api/AmplitudeClient.java
+++ b/src/main/java/com/amplitude/api/AmplitudeClient.java
@@ -2395,7 +2395,7 @@ public class AmplitudeClient {
     }
 
     protected DeviceInfo initializeDeviceInfo() {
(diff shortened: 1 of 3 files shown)


justin-fiedler:
LGTM. Thanks @falconandy!

justin-fiedler (approved):

PR #375, opened 2023-07-06
fix: try to reduce sqlite cursor memory usage

falconandy:
https://github.com/amplitude/Amplitude-Android/blob/main/CHANGELOG.md#2100-oct-12-2016 

```
## 2.10.0 (Oct 12, 2016)
Catch and handle CursorWindowAllocationException thrown when the SDK is querying from > the SQLite DB when app memory is low.
If the exception is caught during initialize, then it is treated as if initialize was > never called.
If the exception is caught during the uploading of unsent events, then the upload is deferred to a later time.
```

https://github.com/amplitude/Amplitude-Android/blob/main/src/main/java/com/amplitude/api/AmplitudeClient.java#L444 

```
} catch (CursorWindowAllocationException e) {  // treat as uninitialized SDK
      logger.e(TAG, String.format(
              "Failed to initialize Amplitude SDK due to: %s", e.getMessage()
      ));
      client.apiKey = null;
}
````

https://github.com/amplitude/Amplitude-Android/blob/main/src/main/java/com/amplitude/api/AmplitudeClient.java#L2187

```
} catch (CursorWindowAllocationException e) {
     // handle CursorWindowAllocationException when fetching events, defer upload
     uploadingCurrently.set(false);
      logger.e(TAG, String.format(
         "Caught Cursor window exception during event upload, deferring upload: %s",
          e.getMessage()
      ));
}
```

New logic was added recently (`IdentifyInterceptor.transferInterceptedIdentify`) that can throw `CursorWindowAllocationException` - the exception is not caught/handled in this case. I believe, we can’t defer `transferInterceptedIdentify` call (similar to `If the exception is caught during the uploading of unsent events, then the upload is deferred to a later time`).

To reduce cursor memory usage now each event is loaded from sqlite database with separate query/cursor. Hope it helps.

Currently my environment doesn't work with Amplitude-Android (build/test) - so I didn't test properly (CI tests only).

Some unused code is removed.

### DIFF
--- a/src/main/java/com/amplitude/api/DatabaseHelper.java
+++ b/src/main/java/com/amplitude/api/DatabaseHelper.java
@@ -338,6 +338,57 @@ class DatabaseHelper extends SQLiteOpenHelper {
     }
 
     protected synchronized List<JSONObject> getEventsFromTable(
+            String table, long upToId, long limit) throws JSONException {
+        try {
+            return getEventsBatchFromTable(table, upToId, limit);
+        } catch (CursorWindowAllocationException e) {
+            return getEventsRowByRowFromTable(table, upToId, limit);
+        }
+    }
+
+    private List<JSONObject> getEventsBatchFromTable(
+            String table, long upToId, long limit) throws JSONException {
+        List<JSONObject> events = new LinkedList<JSONObject>();
+        Cursor cursor = null;
+        try {
+            SQLiteDatabase db = getReadableDatabase();
+            cursor = queryDb(
+                db, table, new String[] { ID_FIELD, EVENT_FIELD },
+                upToId >= 0 ? ID_FIELD + " <= " + upToId : null, null, null, null,
+                ID_FIELD + " ASC", limit >= 0 ? "" + limit : null
+            );
+
+            while (cursor.moveToNext()) {
+                long eventId = cursor.getLong(0);
+                String event = cursor.getString(1);
+                if (Utils.isEmptyString(event)) {
+                    continue;
+                }
+
+                JSONObject obj = new JSONObject(event);
+                obj.put("event_id", eventId);
+                events.add(obj);
+            }
+        } catch (SQLiteException e) {
+            logger.e(TAG, String.format("getEvents from %s failed", table), e);
+            delete();
+        } catch (StackOverflowError e) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/main/java/com/amplitude/api/DatabaseHelper.java [hunk 1]
```diff
@@ -377,7 +369,55 @@ protected synchronized List<JSONObject> getEventsFromTable(
...
+        } finally {
+            close();
+        }
+    }
+
+    protected synchronized JSONObject getEventFromTable(String table, long eventId) throws JSONException {
+        JSONObject event = null;
+        Cursor cursor = null;
+        try {
+            SQLiteDatabase db = getReadableDatabase();
```
justin-fiedler, line 391: @falconandy do you think the extra DB setup/teardown could have other affects on performance?

Since we have the other areas with this issue handled, I would prefer to keep the previous `getEventsFromTable()` logic for the common case, and use this as a fallback for errors in [getTransferIdentifyEvent](https://github.com/amplitude/Amplitude-Android/blob/main/src/main/java/com/amplitude/api/IdentifyInterceptor.java#L118). We could make a new `dbHelper.getIdentifyInterceptorsLowMemory()` and use it on CursorWindow error.

I agree, I don't think we can delay the identify information. Maybe if intercepted identify events fail to load with low memory version we should drop them from the DB? wdyt?

justin-fiedler (commented):
Thanks @falconandy. See questions.

src/main/java/com/amplitude/api/DatabaseHelper.java
falconandy: @justin-fiedler 
> do you think the extra DB setup/teardown could have other affects on performance?

Yes, it could add minor slowness. I've modified code to use original `getEvents` implementation and to switch to fallback `row-by-row` load if `CursorWindowAllocationException` raises. I think, it could help for events, identifies and intercepted identifies (instead of deferring).

>  Maybe if intercepted identify events fail to load with low memory version we should drop them from the DB?

I believe the fallback `row-by-row` schema should help in most cases except a single intercepted identify with huge payload. Yes, we could drop "huge" intercepted identifies as a last resort (with additional logging).

justin-fiedler (approved):

PR #382, opened 2023-07-15
fix: try to correctly handle session change for race condition

justin-fiedler:
# Description

We are [still getting long sessions for some customers in Flutter on Android](https://amplitude.atlassian.net/browse/AMP-76734?focusedCommentId=198409) after the other session fixes.

This change aims to handle the hypothetical case where an event is tracked before the `runOnLogThread(new Runnable()` executes.

That scenario would lead the event being "in foreground" which would extend the existing session.

Maybe it is also possible to move setting `inForeground` to the first line of the onEnterForeground Runnable? wdyt @falconandy?

```diff
void onEnterForeground(final long timestamp) {
-        inForeground = true;
        runOnLogThread(new Runnable() {
            @Override
            public void run() {
+             inForeground = true;
```

https://github.com/amplitude/Amplitude-Android/pull/362/files

### DIFF
--- a/src/main/java/com/amplitude/api/AmplitudeClient.java
+++ b/src/main/java/com/amplitude/api/AmplitudeClient.java
@@ -181,6 +181,7 @@ public class AmplitudeClient {
     private boolean usingForegroundTracking = false;
     private boolean trackingSessionEvents = false;
     private boolean inForeground = false;
+    private boolean isEnteringForeground = false;
     private boolean flushEventsOnClose = true;
     private String libraryName = Constants.LIBRARY;
     private String libraryVersion = Constants.VERSION;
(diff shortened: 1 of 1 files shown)


falconandy (dismissed):
Hi @justin-fiedler.
Earlier I moved `inForeground` setter/getter from log thread body to main thread to avoid possible data races. Now new data race with `isEnteringForeground` is added, I believe.  
Currently I do not see how the new flag can help - could you please provide an example with details to illustrate incorrect/fixed behavior?
Maybe it makes sense to add some internal session-related metrics to event payload (for example, when time gap between events is too large) to help debug session issues.

qingzhuozhen (approved):

justin-fiedler:
> Hi @justin-fiedler. Earlier I moved `inForeground` setter/getter from log thread body to main thread to avoid possible data races. Now new data race with `isEnteringForeground` is added, I believe. Currently I do not see how the new flag can help - could you please provide an example with details to illustrate incorrect/fixed behavior? Maybe it makes sense to add some internal session-related metrics to event payload (for example, when time gap between events is too large) to help debug session issues.

Thanks @falconandy I reviewed with @qingzhuozhen and we feel that this change shouldn't cause any new problems and is worth trying to see if it resolves the issue.

We are in process adding [diagnostics](https://amplitude.atlassian.net/wiki/spaces/GOV/pages/2140831761) to the SDKs to make it easier to locate issues in the near future.

PR #402, opened 2024-06-10
fix: catch all exceptions when trying to intercept identifies

izaaz:
Catch all exceptions when trying to intercept identifies


src/main/java/com/amplitude/api/IdentifyInterceptor.java
```diff
@@ -105,7 +105,7 @@ private JSONObject getTransferIdentifyEvent() {
             identifyEvent.getJSONObject("user_properties").put(Constants.AMP_OP_SET, identifyEventUserProperties);
             dbHelper.removeIdentifyInterceptors(lastIdentifyInterceptorId);
             return identifyEvent;
-        } catch (JSONException e) {
+        } catch (Exception e) {
```
justin-fiedler, line 108: I have started using `Throwable` which covers `Exception`, `Error` and `Violation` - https://developer.android.com/reference/java/lang/Throwable

```suggestion
        } catch (Throwable e) {
```

justin-fiedler (approved):

